PR #39, opened 2024-10-01
Closes #37 - Improve logging in LdapClient

### LINKED ISSUE #37: Improve logging in LdapClient
As of now there is only a very basic logging in LdapClient. This makes the debugging of problems that come from LDAP very difficult. The goal should be to log enough information to reconstruct and debug each step of a communication with a LDAP server, without having to know the specifics/configuration of the server in question.

CRoberto1926:

---

- [x] Sonarcloud link : https://sonarcloud.io/summary/new_code?id=CRoberto1926_kadai&branch=feature%2Fissue-37-ldap-client-logging
- [x] No Release Notes needed
- [ ] Release Notes :

```

```

### DIFF
--- a/rest/kadai-rest-spring/src/main/java/io/kadai/common/rest/ldap/LdapClient.java
+++ b/rest/kadai-rest-spring/src/main/java/io/kadai/common/rest/ldap/LdapClient.java
@@ -44,7 +44,6 @@ import java.util.stream.Stream;
 import javax.naming.InvalidNameException;
 import javax.naming.directory.SearchControls;
 import javax.naming.ldap.LdapName;
-import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.core.env.Environment;
@@ -64,7 +63,8 @@ import org.springframework.stereotype.Component;
 @Component
 public class LdapClient {
 
-  private static final Logger LOGGER = LoggerFactory.getLogger(LdapClient.class);
+  private static final LdapClientLogger LOGGER =
+      new LdapClientLogger(LoggerFactory.getLogger(LdapClient.class));
   private static final String CN = "cn";
 
   private final KadaiConfiguration kadaiConfiguration;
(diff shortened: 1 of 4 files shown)


common/kadai-common/src/main/java/io/kadai/common/internal/util/LogSanitizer.java
```diff
@@ -18,6 +18,7 @@
 
 package io.kadai.common.internal.util;
 
+// TODO FIXME replace with OWASP SANITIZER: https://github.com/OWASP/owasp-java-encoder
```
holgerhagen: I suggest we rather open an issue than create a TODO comment.

rest/kadai-rest-spring/src/main/java/io/kadai/common/rest/ldap/LdapClient.java
```diff
@@ -840,6 +848,179 @@ void testMinSearchForLength(final String name) throws InvalidArgumentException {
     }
   }
 
+  private AccessIdRepresentationModel lookupAndDebugLog(
```
holgerhagen: The fact that we write debug logs should not be in the operation name, shouldn't it? It is rather a side effect depending on the log level.

holgerhagen (changes requested):
Please add guarded logging to all statements (if isDebugEnabled...)

common/kadai-common/src/main/java/io/kadai/common/internal/util/LogSanitizer.java
CRoberto1926: Done: https://github.com/kadai-io/kadai/issues/69

rest/kadai-rest-spring/src/main/java/io/kadai/common/rest/ldap/LdapClient.java
CRoberto1926: I prefer to be as precise as possible in a method name. In this was the reader doesn't have to read the method definition. But you also have a point, so I removed "AndDebugLog" from the "search" and "lookup" methods.

PR #158, opened 2024-11-13
Closes #93 - improve the pipeline

### LINKED ISSUE #93: Split the "deploy artifacts to Sonatype" action, so that the "update Version to SNAPSHOT" step is extra
At the moment, if the "update Version to SNAPSHOT" step fails, then all parts of "deploy artifacts to Sonatype" need to be rerun. That's a problem, because in this case we need to delete the deployment from Sonatype

ryzheboka:
updating Kadai version to next Snapshot is now executed as an extra job instead of just being a step

---

- [x] No Release Notes needed
- [ ] Release Notes :

```

```

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -35,7 +35,7 @@ env:
 jobs:
   compile_backend:
     name: Compile all maven modules
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     steps:
       - name: Git checkout
         uses: actions/checkout@v4
@@ -74,7 +74,7 @@ jobs:
 
   compile_frontend:
     name: Compile kadai-web
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     steps:
       - name: Git checkout
         uses: actions/checkout@v4
@@ -129,7 +129,7 @@ jobs:
         uses: andymckay/cancel-action@0.5
 
   test_frontend:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     name: Test kadai-web
     needs: [ compile_frontend ]
     steps:
@@ -173,7 +173,7 @@ jobs:
         uses: andymckay/cancel-action@0.5
 
   test_e2e:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     name: Test E2E
     needs: [ compile_frontend, compile_backend ]
     steps:
@@ -243,7 +243,7 @@ jobs:
         uses: andymckay/cancel-action@0.5
 
   test_backend:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     name: Test ${{ matrix.module }} on ${{ matrix.database }}
     needs: [ compile_backend ]
     strategy:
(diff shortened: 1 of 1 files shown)


.github/workflows/continuous-integration.yml
```diff
@@ -392,6 +392,16 @@ jobs:
           GPG_PASSPHRASE: ${{ secrets.GPG_PASSPHRASE }}
           MAVEN_USERNAME: ${{ secrets.SONATYPE_USERNAME }}
           MAVEN_PASSWORD: ${{ secrets.SONATYPE_PASSWORD }}
+      - name: Cancel workflow
+        if: failure()
+        uses: andymckay/cancel-action@0.5
```
CRoberto1926, line 397: is this also going to revert the changes from change_version.sh?

.github/workflows/continuous-integration.yml
```diff
@@ -392,6 +392,16 @@ jobs:
...
+
+  update_version:
+    runs-on: ubuntu-22.04
+    name: Update version to next snapshot and push back
+    if: github.repository == 'kadai-io/kadai' && startsWith(github.ref, 'refs/tags/v') && github.head_ref == ''
+    needs: [ test_frontend, test_e2e, test_backend ]
```
CRoberto1926: I would say needs: [release_artifacts] or? Would that work?

.github/workflows/continuous-integration.yml
```diff
@@ -392,6 +392,16 @@ jobs:
           GPG_PASSPHRASE: ${{ secrets.GPG_PASSPHRASE }}
           MAVEN_USERNAME: ${{ secrets.SONATYPE_USERNAME }}
           MAVEN_PASSWORD: ${{ secrets.SONATYPE_PASSWORD }}
+      - name: Cancel workflow
+        if: failure()
+        uses: andymckay/cancel-action@0.5
+
+  update_version:
+    runs-on: ubuntu-22.04
```
CRoberto1926, line 400: This is not part of your PR, but I would suggest to use everywhere the same ubuntu version. The last LTS should be 24.04, but we have to check if everything works. Could you write an issue for this please?

.github/workflows/continuous-integration.yml
ryzheboka: As change_version.sh doesn't commit anything or upload any artifacts, I see no reason why it wouldn't be reverted. How could we check this for sure?

.github/workflows/continuous-integration.yml
ryzheboka: Yes, I could introduce a constant for the Ubuntu version and set it to 22.04, as we know that it works. I can also create an issue for the update to 24.04

.github/workflows/continuous-integration.yml
ryzheboka: Yes :)

.github/workflows/continuous-integration.yml
CRoberto1926: mvn versions:commit is not committing anything?

.github/workflows/continuous-integration.yml
CRoberto1926: I would be in favor of a constant for ubuntu-22.04, yes

.github/workflows/continuous-integration.yml
ryzheboka: As can be seen in our repo history, after a release, we have one automatical commit, for example the "[Updated poms to version 9.1.1-SNAPSHOT](https://github.com/kadai-io/kadai/commit/24090bd351d365d9a2adc207776c7003e1dcba9f)" commit for last update. This commit is done in the next step in ci/commitPoms.sh. The step that I added the cacelation to was previously also there, but produced no additional commit. According to the [doc](https://www.mojohaus.org/versions/versions-maven-plugin/commit-mojo.html), mvn versions:commit doesn't commit anything in the classical sense. However, it still would be better to prove or try out if the changes by the change_version scripts are revertable by the cancel actions.

.github/workflows/continuous-integration.yml
CRoberto1926: Can you change that? Or do you have counter-arguments?

CRoberto1926 (approved):

PR #159, opened 2024-11-13
Closes #73 - add routing with owner

### LINKED ISSUE #73: extend TaskRoutingProvider so it can be used to set the owner

ryzheboka:

---

- [x] No Release Notes needed
- [ ] Release Notes :

```

```

### DIFF
--- a/lib/kadai-core-test/src/test/java/acceptance/taskrouting/TaskRoutingAccTest.java
+++ b/lib/kadai-core-test/src/test/java/acceptance/taskrouting/TaskRoutingAccTest.java
@@ -21,15 +21,17 @@ package acceptance.taskrouting;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.AssertionsForClassTypes.assertThatThrownBy;
 
-import acceptance.taskrouting.TaskRoutingAccTest.TaskRoutingProviderForDomainA;
+import io.kadai.KadaiConfiguration.Builder;
 import io.kadai.classification.api.ClassificationService;
 import io.kadai.classification.api.models.ClassificationSummary;
 import io.kadai.common.api.KadaiEngine;
 import io.kadai.common.api.exceptions.InvalidArgumentException;
+import io.kadai.spi.routing.api.RoutingTarget;
 import io.kadai.spi.routing.api.TaskRoutingProvider;
 import io.kadai.task.api.TaskService;
 import io.kadai.task.api.models.Task;
 import io.kadai.testapi.DefaultTestEntities;
+import io.kadai.testapi.KadaiConfigurationModifier;
 import io.kadai.testapi.KadaiInject;
 import io.kadai.testapi.KadaiIntegrationTest;
 import io.kadai.testapi.WithServiceProvider;
@@ -38,17 +40,15 @@ import io.kadai.testapi.security.WithAccessId;
 import io.kadai.workbasket.api.WorkbasketPermission;
 import io.kadai.workbasket.api.WorkbasketService;
 import io.kadai.workbasket.api.models.WorkbasketSummary;
+import java.util.Optional;
 import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Nested;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestInstance;
+import org.junit.jupiter.api.TestInstance.Lifecycle;
 
-@WithServiceProvider(
-    serviceProviderInterface = TaskRoutingProvider.class,
-    serviceProviders = TaskRoutingProviderForDomainA.class)
 @KadaiIntegrationTest
 class TaskRoutingAccTest {
-
-  @KadaiInject KadaiEngine kadaiEngine;
-  @KadaiInject TaskService taskService;
 
   ClassificationSummary classificationSummary;
   WorkbasketSummary domainAWorkbasket;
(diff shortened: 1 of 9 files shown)


lib/kadai-core-test/src/test/resources/fullKadai.properties
```diff
@@ -1,6 +1,7 @@
 # general configuration
 kadai.domains=DOMAIN_A | DOMAIN_B
 kadai.servicelevel.validation.enforce=false
+kadai.routing.includeOwner=true
```
holgerhagen, line 4: the defautl should be false, right? And shouldn't this contain the default?

lib/kadai-core-test/src/test/java/acceptance/taskrouting/TaskRoutingAccTest.java
```diff
@@ -74,41 +73,174 @@ void setUp(ClassificationService classificationService, WorkbasketService workba
...
+      @WithAccessId(user = "user-1-1")
+      @Test
+      void should_SetWorkbasketAndKeepOwnerForTask_When_TaskRouterDeterminesRoutingTarget()
+          throws Exception {
+        Task task = taskService.newTask(null, "DOMAIN_A");
+        task.setOwner("user-1-2");
+        task.setClassificationKey(classificationSummary.getKey());
+        task.setPrimaryObjRef(DefaultTestEntities.defaultTestObjectReference().build());
+
+        Task createdTask = taskService.createTask(task);
```
holgerhagen, line 237: Just a question: how do I find out, which RoutingProvider is used?

lib/kadai-core/src/main/java/io/kadai/spi/routing/api/TaskRoutingProvider.java
```diff
@@ -29,6 +29,10 @@
  */
 public interface TaskRoutingProvider {
 
+  static RoutingTarget newEmptyRoutingTarget() {
```
holgerhagen: Do we need this? Why not using "new RoutingTarget()"?

lib/kadai-core/src/main/java/io/kadai/spi/routing/api/TaskRoutingProvider.java
```diff
@@ -65,5 +69,38 @@ public interface TaskRoutingProvider {
    * @param task the {@linkplain Task} for which a {@linkplain Workbasket} must be determined
    * @return the {@linkplain Workbasket#getId() id} of the {@linkplain Workbasket}
    */
-  String determineWorkbasketId(Task task);
+  default String determineWorkbasketId(Task task) {
+    return null;
+  }
+
+  /**
+   * This method is used instead of {@linkplain TaskRoutingProvider#determineWorkbasketId(Task)} if
```
holgerhagen, line 74: Shoudl we add a little bit more details here? Like "Use this if you want to initialize the owner during Routing"

And didn't you have a test that the owner is not overwritten if existing? Maybe we should specify this behaviour as well?

lib/kadai-core/src/main/java/io/kadai/spi/routing/internal/TaskRoutingManager.java
```diff
@@ -90,6 +91,42 @@ public String determineWorkbasketId(Task task) {
...
   }
 
+  /**
+   * Determines a {@linkplain RoutingTarget} for a given {@linkplain Task}.
+   * Algorithm: see {@linkplain TaskRoutingManager#determineWorkbasketId(Task)}.
+   *
+   * @param task the {@linkplain Task} for which a {@linkplain RoutingTarget} is to be determined
+   * @return the {@linkplain RoutingTarget} for the creation of the {@linkplain Task}
+   */
+  public RoutingTarget determineRoutingTarget(Task task) {
```
holgerhagen: I'm curious: what happens if we have an old RoutingProvider which implements the old interface with only one operation? Do we need versioning?

holgerhagen (changes requested):

lib/kadai-core-test/src/test/resources/fullKadai.properties
ryzheboka: No, fullKadai.properties always contains a value different from default for testing purposes. For example, see kadai.servicelevel.validation.enforce=false. Der default für kadai.servicelevel.validation.enforce ist true

lib/kadai-core-test/src/test/java/acceptance/taskrouting/TaskRoutingAccTest.java
ryzheboka: It's written above the class declaration next to "serviceProviders = "

lib/kadai-core/src/main/java/io/kadai/spi/routing/api/TaskRoutingProvider.java
ryzheboka: We need this for users, so that they can create a RoutingTarget without using our RoutingTarget class. However, I would be open to let our users also use new RoutingTarget() and delete this method. Let's discuss this!

lib/kadai-core/src/main/java/io/kadai/spi/routing/api/TaskRoutingProvider.java
ryzheboka: I will add this information :)

lib/kadai-core/src/main/java/io/kadai/spi/routing/internal/TaskRoutingManager.java
ryzheboka: This method only gets called if kadai.routing.includeOwner=true is set. In this case, only the method determineRoutingTarget of the RoutingProvider is relevant. If the RoutingProvider doesn't implement such method, then the default method implementation is called. The default method implementation returns an empty RoutingTarget. If there is no other RoutingProvider returning a nonNull RoutingTarget, a suitable error is thrown. So I don't see a problem here. Would you agree?

lib/kadai-core-test/src/test/resources/fullKadai.properties
holgerhagen: ok

holgerhagen (approved):

PR #242, opened 2024-11-27
Closes #241: Filter Logs of KadaiConfiguration

### LINKED ISSUE #241: Filter Logs of KadaiConfiguration
In the log displaying all properties of Kadai Configuration, we only want to log kadai.* properties and also filter out any passwords or sensitive properties

jamesrdi:

---

- [ ] I put the ticket or multiple tickets in review
- [x] Sonarcloud link : [\<add the link here>](https://sonarcloud.io/summary/new_code?id=jamesrdi_kadai&branch=KAD-241)
- [ ] No Release Notes needed
- [ ] Release Notes :

```

```

### DIFF
--- a/lib/kadai-core-test/src/test/java/acceptance/KadaiEngineInitializationTest.java
+++ b/lib/kadai-core-test/src/test/java/acceptance/KadaiEngineInitializationTest.java
@@ -20,7 +20,10 @@ package acceptance;
 
 import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
 
+import io.kadai.KadaiConfiguration;
+import io.kadai.KadaiConfiguration.Builder;
 import io.kadai.common.api.KadaiEngine;
+import io.kadai.common.api.KadaiEngine.ConnectionManagementMode;
 import io.kadai.common.internal.KadaiEngineImpl;
 import io.kadai.spi.priority.api.PriorityServiceProvider;
 import io.kadai.spi.priority.internal.PriorityServiceManager;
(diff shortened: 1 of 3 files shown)


lib/kadai-core/src/main/java/io/kadai/common/internal/KadaiEngineImpl.java
```diff
@@ -128,14 +131,34 @@ public class KadaiEngineImpl implements KadaiEngine {
...
+  }
+
   protected KadaiEngineImpl(
       KadaiConfiguration kadaiConfiguration,
       ConnectionManagementMode connectionManagementMode,
       TransactionFactory transactionFactory)
       throws SQLException {
+    Builder builder = new Builder(kadaiConfiguration);
+    builder.setProperties(filterKadaiProperties(kadaiConfiguration.getProperties()));
+    KadaiConfiguration copiedKadaiConfiguration = builder.build();
```
CRoberto1926: Why do you need to instantiate a whole new `KadaiConfiguration`?

Isn't it enough to just modify the `KadaiConfiguration.toString()`method to filter out the properties?

grafik

jamesrdi: But the toString method can potentially be used by other methods, that require all properties unfiltered to be given as string

CRoberto1926: But is it now used?

jamesrdi: I don't think it is used right now yet

CRoberto1926: I think that it is safe to assume that changing the output in the toString() method is not dangerous then, what do you think?

lib/kadai-core-test/src/test/java/acceptance/KadaiEngineInitalizationTest.java
```diff
@@ -55,6 +67,46 @@ public OptionalInt calculatePriority(TaskSummary taskSummary) {
     }
   }
 
+  private static class TestKadaiEngineImpl extends KadaiEngineImpl {
```
CRoberto1926: Why do you need this class and can't just call `KadaiEngineImpl.createKadaiEngine`?

lib/kadai-core-test/src/test/java/acceptance/KadaiEngineInitalizationTest.java
```diff
@@ -20,23 +20,35 @@
...
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.TestInstance;
 import org.junit.jupiter.api.TestInstance.Lifecycle;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.springframework.boot.test.system.CapturedOutput;
+import org.springframework.boot.test.system.OutputCaptureExtension;
 
+@ExtendWith(OutputCaptureExtension.class)
 @KadaiIntegrationTest
 public class KadaiEngineInitalizationTest {
```
CRoberto1926: Can you fix the typo here please :)

`KadaiEngineInitializationTest`

CRoberto1926 (changes requested):

lib/kadai-core-test/src/test/java/acceptance/KadaiEngineInitalizationTest.java
jamesrdi: You are right, I did not know we had that method

lib/kadai-core-test/src/test/java/acceptance/KadaiEngineInitalizationTest.java
jamesrdi: Yes I can

CRoberto1926 (approved):

Issue #83, opened 2024-10-22
Test public Map<String, List<String>> searchAccessIdForGroupsAndPermissionsByDn(List<String> dns)


ryzheboka:
The method public Map<String, List<String>> searchAccessIdForGroupsAndPermissionsByDn(List<String> dns) is not well tested

CRoberto1926:
We should ask the users if they are using it. Otherwise we can delete it.

PR #393, opened 2025-01-10
Closes #357 - add a new ldap attribute for group access id

### LINKED ISSUE #357: Extend the ldap attributes so that there is an attribute for the group name and an attribute for the group acces id
Right now, we only have the attribute for group name. This lowercase value of this attribute is also used to create the access id of the group. We need an additional attribute for the access id of the group

ryzheboka:
The test coverage in the sonarcloud evaluation is not accurate, as it doesn't recognize that some tests run with a configuration different from the default. Additionaly, there is one untested method in the LdapClient class that is currently discussed in #83.

---

- [x] Sonarcloud link : https://sonarcloud.io/summary/new_code?id=ryzheboka_kadai&branch=KAD-357 
- [ ] No documentation update needed
- [x] Link to PR with documentation update: https://github.com/kadai-io/kadai-doc/pull/45
- [x] No Release Notes needed
- [ ] Release Notes :

```

```

### DIFF
--- a/rest/kadai-rest-spring-test-lib/src/main/resources/kadai-test.ldif
+++ b/rest/kadai-rest-spring-test-lib/src/main/resources/kadai-test.ldif
@@ -394,6 +394,7 @@ uniquemember: uid=user-2-2,cn=users,OU=Test,O=KADAI
 cn: ksc-users
 objectclass: groupofuniquenames
 objectclass: top
+gid: ksc-users
 
 dn: cn=ksc-teamleads,cn=groups,OU=Test,O=KADAI
 uniquemember: uid=teamlead-1,cn=users,OU=Test,O=KADAI
@@ -401,6 +402,7 @@ uniquemember: uid=teamlead-2,cn=users,OU=Test,O=KADAI
 cn: ksc-teamleads
 objectclass: groupofuniquenames
 objectclass: top
+gid: ksc-teamleads
 
 dn: cn=business-admins,cn=groups,OU=Test,O=KADAI
 uniquemember: uid=teamlead-1,cn=users,OU=Test,O=KADAI
@@ -416,6 +418,7 @@ uniquemember: uid=teamlead-2,cn=users,OU=Test,O=KADAI
 cn: monitor-users
 objectclass: groupofuniquenames
 objectclass: top
+gid: monitor-users-ID
 
 ########################
 # Permissions
@@ -428,6 +431,7 @@ permission: Kadai:CallCenter:AB:AB/A:CallCenter
 cn: g01
 objectclass: groupofuniquenames
 objectclass: top
+pid: Kadai:CallCenter:AB:AB/A:CallCenter
 
 dn: cn=g02,cn=groups,OU=Test,O=KADAI
 uniquemember: uid=user-1-2,cn=users,OU=Test,O=KADAI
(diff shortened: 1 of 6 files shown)


rest/kadai-rest-spring/src/main/java/io/kadai/common/rest/ldap/LdapClient.java
```diff
@@ -107,7 +107,7 @@ public List<AccessIdRepresentationModel> searchUsersAndGroupsAndPermissions(fina
       }
     } else {
       accessIds.addAll(searchUsersByNameOrAccessId(name));
-      accessIds.addAll(searchGroupsByName(name));
+      accessIds.addAll(searchGroupsByNameOrAccessId(name));
       accessIds.addAll(searchPermissionsByName(name));
```
holgerhagen: shouldn't this be the same as the groups?

holgerhagen (changes requested):

rest/kadai-rest-spring/src/main/java/io/kadai/common/rest/ldap/LdapClient.java
ryzheboka: The access id of a permission should always be the permission name in lowercase according to the requirements. That's why we don't need to search permissions by access id.

rest/kadai-rest-spring/src/main/java/io/kadai/common/rest/ldap/LdapClient.java
```diff
@@ -799,20 +833,20 @@ String[] getLookUpUserInfoAttributesToReturn() {
...
-      getUserMobilePhoneAttribute(),
-      getUserEmailAttribute(),
-      getUserOrgLevel1Attribute(),
-      getUserOrgLevel2Attribute(),
-      getUserOrgLevel3Attribute(),
-      getUserOrgLevel4Attribute()
+    return new String[]{
+        getUserIdAttribute(),
+        getUserMemberOfGroupAttribute(),
+        getUserPermissionsAttribute(),
```
holgerhagen, line 839: what is this for? Didn't we replace it with permission name?

rest/kadai-rest-spring-test-lib/src/main/resources/kadai-test.ldif
```diff
@@ -428,13 +431,24 @@ permission: Kadai:CallCenter:AB:AB/A:CallCenter
...
 objectclass: top
+pid: Kadai:CallCenter:AB:AB/A:CallCenter
 
 dn: cn=g02,cn=groups,OU=Test,O=KADAI
 uniquemember: uid=user-1-2,cn=users,OU=Test,O=KADAI
 permission: Kadai:CallCenter:AB:AB/A:CallCenter-vip
 cn: g02
 objectclass: groupofuniquenames
 objectclass: top
+pid: Kadai:CallCenter:AB:AB/A:CallCenter-vip
```
holgerhagen, line 442: maybe this should be different to "permission" in the tests?

holgerhagen (changes requested):

rest/kadai-rest-spring/src/main/java/io/kadai/common/rest/ldap/LdapClient.java
ryzheboka: These are the attributes of the user, not the permission. That's why it should be UserPermissionAttribute. Similarly, UserMemeberOfGroupAttribute is used, not the GroupNameAttribute.

rest/kadai-rest-spring-test-lib/src/main/resources/kadai-test.ldif
ryzheboka: It has to be the same for these two permissions, because they are used in the tests that are run on both setups, the setup with permissionIdAttribute and the setup without the attribute. In order for these tests to pass, the values of the two attributes have to be the same. That's why there is a new permission that is not part of the other test. The new permission has different values

holgerhagen (approved):

PR #65, opened 2024-10-09
Closes #64 - stop deploying SNAPSHOTS in the pipeline

ryzheboka:
### Thanks for your PR! Please fill out the following list :)

---

- [x] I put the ticket or multiple tickets in review
- [x] Commit message format → Closes #&lt;Issue Number&gt; - Your commit message.
- [ ] Sonarcloud link : https://sonarcloud.io/summary/new_code?id=ryzheboka_kadai&branch=KAD-64
- [x] No documentation update needed
- [ ] Link to PR with documentation update: \<add the link here>
- [x] No Release Notes needed
- [ ] Release Notes :

```

```

PR #510, opened 2025-01-30
Closes #437 - Revert "Closes #64 - stop deploying SNAPSHOTS in the pipeline"

### LINKED ISSUE #64: Remove SNAPSHOT Deployment from the Pipeline

### LINKED ISSUE #437: Restore SNAPSHOT Deployment in the pipeline
With https://github.com/kadai-io/kadai/issues/64 the SNAPSHOT Deployment was removed from the pipeline, because it was not supported from Sonatype Maven Central Repository. With the last version of the Maven Plugin this possibility is supported again (https://central.sonatype.org/publish/publish-portal-maven/#070).

SNAPSHOT Deployment should be restored in KADAI.

bruderj15:
Reverts kadai-io/kadai#65

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -333,7 +333,7 @@ jobs:
   release_artifacts:
     runs-on: ubuntu-22.04
     name: Release artifacts to OSS Sonatype
-    if: github.repository == 'kadai-io/kadai' && startsWith(github.ref, 'refs/tags/v') && github.head_ref == ''
+    if: github.repository == 'kadai-io/kadai' && ( startsWith(github.ref, 'refs/tags') || github.ref == 'refs/heads/master' ) && github.head_ref == ''
     needs: [ test_frontend, test_e2e, test_backend ]
     # as documented in the gpg manual (https://www.gnupg.org/documentation/manuals/gnupg/Invoking-GPG_002dAGENT.html)
     # we should execute this command before interacting with gpg (otherwise gpg won't work)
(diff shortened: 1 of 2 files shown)


.github/workflows/continuous-integration.yml
```diff
@@ -333,7 +333,7 @@ jobs:
   release_artifacts:
     runs-on: ubuntu-22.04
     name: Release artifacts to OSS Sonatype
-    if: github.repository == 'kadai-io/kadai' && startsWith(github.ref, 'refs/tags/v') && github.head_ref == ''
```
CRoberto1926, line 336: I think we can leave the `/v` there, because it checks more precisely for versions. What do you think?

CRoberto1926:
The same should be also done for KadaiAdapter: https://github.com/kadai-io/KadaiAdapter/pull/19/files

CRoberto1926 (commented):
Can we test this changes?

Do we have to change something?

https://central.sonatype.org/publish/publish-portal-snapshots/

.github/workflows/continuous-integration.yml
bruderj15: Seems redundant as all tags are versioned releases here, will add for readability

bruderj15:
> Do we have to change something?

According to provided documentation, done changes shall be sufficient.

> Can we test this changes?

Barely. We could:
- test locally - requires setup,
- temporarily adjust job to be triggered here on PR or
- merge into master and hope for the best (, whew).

All three shall result in actual snapshot deployments to the snapshot repository.

EDIT: We need to __enable Snapshot-Deployment__ for our namespace, see [this](https://central.sonatype.org/publish/publish-portal-snapshots/#enabling-snapshot-releases-for-your-namespace)

CRoberto1926 (approved):
LGTM

Test on master? 🏃

bruderj15:
@CRoberto1926 
Sonatypes Snapshot-Repository responded with a `403`, see [this](https://github.com/kadai-io/kadai/actions/runs/13199971936/job/36849882726#step:9:3439).

> EDIT: We need to enable Snapshot-Deployment for our namespace, see [this](https://central.sonatype.org/publish/publish-portal-snapshots/#enabling-snapshot-releases-for-your-namespace)

Is it enabled?

PR #530, opened 2025-02-06
Closes #512 - Test KadaiEngine with different ConnectionManagementModes

### LINKED ISSUE #512: Test KadaiEngine with different `ConnectionManagementMode`s
Following #44 and the discussion in https://github.com/kadai-io/kadai/pull/491#issuecomment-2627302190, revisit the old Taskana-Tests, transfer those worth to Kadai and extend where necessary.

bruderj15:

---

- [ ] I put the ticket or multiple tickets in review
- [x] No Release Notes needed
- [ ] Release Notes :

```

```


bruderj15:
`ConnectionManagementMode`=`EXPLICIT` succeeds with `H2` and `Postgres` but fails with `DB2`, see [CI](https://github.com/bruderj15/kadai-fork/actions/runs/13179067278/job/36785314118#step:7:3527).

Seems like a bug with `DB2`. 
Having no knowledge about `DB2` at all, I cannot find it.

CRoberto1926 (commented):
You could probably summarize `KadaiEngineAutocommitModeTest`, `KadaiEngineExplicitModeTest` and `KadaiEngineParticipateModeTest` in one `KadaiEngineModesTest` class, and then group them using `@Nested`. I think they would be better grouped together.

bruderj15:
Wondering big time here.
Current failures are:
- `EXPLICIT` succeeds with `H2` and `Postgres` but fails with `DB2`, see [CI](https://github.com/bruderj15/kadai-fork/actions/runs/13179067278/job/36785314118#step:7:3527)
- ~`PARTICIPATE` fails completely, see [CI](https://github.com/kadai-io/kadai/actions/runs/13333478756/job/37243261060#step:7:3574)~

bruderj15:
@jamesrdi you are familiar with the core, right? What's going wrong here?

lib/kadai-core-test/src/test/java/acceptance/common/KadaiEngineModesTest.java
```diff
@@ -0,0 +1,235 @@
...
+      KadaiConfiguration kadaiConfiguration =
+          new KadaiConfiguration.Builder(dataSource, true, schemaName, false)
+              .initKadaiProperties()
+              .build();
+      thisKadaiEngineProxy =
+          new KadaiEngineProxy(
+              KadaiEngine.buildKadaiEngine(
+                  kadaiConfiguration, ConnectionManagementMode.PARTICIPATE));
+      thisKadaiEngineProxy.getSqlSession().getConfiguration().addMapper(TestUserMapper.class);
+      transactionTemplate = new TransactionTemplate(new DataSourceTransactionManager(dataSource));
```
jamesrdi, line 48: For PARTICIPATE mode, I believe the issue lies in the fact that the kadaiEngine does not detect the transactionTemplate, which is why when calling executeInDatabaseConnection(), kadaiEngine creates a new transaction instead of participating in the transactionTemplate, ignoring the rollback from status.setRollbackOnly()

This is because in the KadaiEngineImpl.java, in createTransactionFactory(), ManagedTransactionFactory was used instead of SpringManagedTransactionFactory. ManagedTransactionFactory cannot detect transactions managed by spring, but in the test, you initialised a transaction with TransactionTemplate, which is managed by Spring.

A solution would be to change ManagedTransactionFactory to SpringManagedTransactionFactory in KadaiEngineImpl class, I have tried it and it worked for the test, but it also fails on other tests since some classes rely on Jakarta and not Spring, like KadaiProducers.java

Or you can initialise a transaction that is managed by jakarta instead of Spring in the test.

lib/kadai-core-test/src/test/java/acceptance/common/KadaiEngineModesTest.java
```diff
@@ -0,0 +1,235 @@
...
+          thisKadaiEngineProxy.getSqlSession().getMapper(TestUserMapper.class);
+      TestUserMapper thatMapper =
+          thatKadaiEngineProxy.getSqlSession().getMapper(TestUserMapper.class);
+      TestUser expected =
+          new TestUser("teamlead-42", "Max", "Mustermann", "Long name of teamlead-42");
+
+      thisKadaiEngineProxy
+          .getEngine()
+          .executeInDatabaseConnection(() -> thisMapper.insert(expected));
+
```
jamesrdi, line 227: I believe you should explicitly add `thisConnection.rollback();` here, since in explicit mode, you should explicitly specify whether to rollback or commit

bruderj15: The test actually wants the changes dangling, neither commited nor rolled back.
Works for Postgres & H2, just DB2 is failing.

lib/kadai-core-test/src/test/java/acceptance/common/KadaiEngineModesTest.java
bruderj15: I will try the latter, thanks for the insight!

lib/kadai-core-test/src/test/java/acceptance/common/KadaiEngineModesTest.java
```diff
@@ -0,0 +1,235 @@
...
+    public void setupKadaiEngines() throws Exception {
+      String schemaName = TestContainerExtension.determineSchemaName();
+      DataSource dataSource = TestContainerExtension.DATA_SOURCE;
+      KadaiConfiguration kadaiConfiguration =
+          new KadaiConfiguration.Builder(dataSource, true, schemaName, false)
+              .initKadaiProperties()
+              .build();
+      thisKadaiEngineProxy =
+          new KadaiEngineProxy(
+              KadaiEngine.buildKadaiEngine(
```
gitgoodjhe: For participate, please give SpringKadaiEngine a try instead of KadaiEngine here

bruderj15: This is working, thank you!
(Makes a whole lot of sense)

bruderj15:
The failing test with `EXPLICIT` for DB2 passes if
- we rollback after insertion (obviously)
- we close the connection after insertion (implies rollback)

Therefore it's obvious that the dangling changes for the transaction in progress are leaked.
I suspect faulty transaction-boundaries.

However:
- `Connection::getTransactionIsolation` is `READ_COMMITED` for `this` and `that` - pointing to correct configuration of sessions
- `Connection::hashCode` is differing for `this` and `that` - pointing to correct usage of sessions

bruderj15:
Also fails for `TRANSACTION_REPEATABLE_READ` and even `TRANSACTION_SERIALIZABLE` - the highest isolation levels.
This proves that it's certainly the DBMS responsible here.

bruderj15:
It's the `DB2_WITH_UR` snippet we always set for DB2...
https://github.com/kadai-io/kadai/blob/790cad891731794b5046c59712a3dfa9f01a918c/common/kadai-common/src/main/java/io/kadai/common/internal/util/SqlProviderUtil.java#L28

UR = UNCOMMITED_READS

Takes precedence over session-level transaction boundaries...

bruderj15:
Closing this in favor of #1346

PR #642, opened 2025-03-03
Closes #549 - Rethrowing checked exceptions in lambdas

### LINKED ISSUE #549: Replacing `Checked`-Family with machinery rethrowing actual exceptions
Javas checked exceptions collide with lambdas.
Sometimes we would want to throw checked exceptions inside a lambda, which does not meet the signature of, e.g. `Function::apply`.
Therefore helpers have been added, e.g. [CheckedFunction](https://github.com/kadai-io/kadai/blob/master/common/kadai-common/src/main/java/io/kadai/common/internal/util/CheckedFunction.java):
```java
@FunctionalInterface
public interface CheckedFunction<T, R, E extends Throwable> {
  static <T, R> Function<T, R> wrap(CheckedFunction<T, R, Throwable> checkedFunction) {
    return t -> {
      try {
        return checkedFunction.apply(t);
      } catch (Throwable e) {
        throw new SystemException("Caught exception", e);
      }
    };
  }
  // ...
  R apply(T t) throws E;
}
```
This rethrows everything as a [SystemException](https://github.com/kadai-io/kadai/blob/7f6a5da5b28a28fa9e9f020cb7fecaab73da3f02/common/kadai-common/src/main/java/io/kadai/common/api/exceptions/SystemException.java#L22), see its specification:
```java

/** This exception is thrown when a generic KADAI problem is encountered. */
public class SystemException extends KadaiRuntimeException {
  public static final String ERROR_KEY = "CRITICAL_SYSTEM_ERROR";
  // ...
}
```

This obviously is not ideal.
I propose to rethrow the actual exception.

Before finding this, I added a fully-tested implementation of it [here](https://github.com/kadai-io/kadai/pull/490/files#diff-446826e8907f00f189e5261827aadbe98bac17e64c5c4a41cd500f1a26af135f).

bruderj15:
## Definition of Done

  - will be squashed on merge

### DIFF
--- a/common/kadai-common-test/src/main/java/io/kadai/common/test/security/JaasExtension.java
+++ b/common/kadai-common-test/src/main/java/io/kadai/common/test/security/JaasExtension.java
@@ -18,7 +18,6 @@
 
 package io.kadai.common.test.security;
 
-import static io.kadai.common.internal.util.CheckedFunction.wrapExceptFor;
 import static org.junit.platform.commons.support.AnnotationSupport.isAnnotated;
 
 import io.kadai.common.api.exceptions.SystemException;
@@ -100,7 +99,15 @@ public class JaasExtension implements InvocationInterceptor, TestTemplateInvocat
     subject.getPrincipals().addAll(getPrincipals(withAccessId));
 
     Function<Invocation<T>, T> proceedInvocation =
-        wrapExceptFor(Invocation::proceed, TestAbortedException.class);
+        f -> {
+          try {
+            return f.proceed();
+          } catch (TestAbortedException e) {
+            throw e;
+          } catch (Throwable e) {
+            throw new SystemException("Caught exception", e);
+          }
+        };
     PrivilegedAction<T> performInvocation = () -> proceedInvocation.apply(invocation);
     return Subject.doAs(subject, performInvocation);
   }
--- a/common/kadai-common/src/test/java/io/kadai/common/internal/util/LambdaCheckedExceptionsTest.java
+++ b/common/kadai-common/src/test/java/io/kadai/common/internal/util/LambdaCheckedExceptionsTest.java
@@ -42,7 +42,7 @@ public class LambdaCheckedExceptionsTest {
 
       final ThrowingCallable call =
           () ->
-              CheckedConsumer.wrap(
+              CheckedConsumer.rethrowing(
                       o -> {
                         throw expected;
                       })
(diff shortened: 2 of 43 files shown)


common/kadai-common-test/src/main/java/io/kadai/common/test/security/JaasExtension.java
```diff
@@ -100,7 +99,15 @@ private static <T> T performInvocationWithAccessId(
     subject.getPrincipals().addAll(getPrincipals(withAccessId));
 
     Function<Invocation<T>, T> proceedInvocation =
-        wrapExceptFor(Invocation::proceed, TestAbortedException.class);
+        f -> {
+          try {
+            return f.proceed();
+          } catch (TestAbortedException e) {
+            throw e;
+          } catch (Throwable e) {
```
jannikFuellgrafEnvite, line 107: Why do you catch a Throwable and not an Exception here?

bruderj15: Because the signature of `Invocation::proceed` demands to:
```java
	/**
	 * An invocation that returns a result and may throw a {@link Throwable}.
	 *
	 * <p>This interface is not intended to be implemented by clients.
	 *
	 * @param <T> the result type
	 * @since 5.5
	 */
	@API(status = STABLE, since = "5.10")
	interface Invocation<T> {

		/**
		 * Proceed with this invocation.
		 *
		 * @return the result of this invocation; potentially {@code null}.
		 * @throws Throwable in case the invocation failed
		 */
		T proceed() throws Throwable;
               
                // ...
	}
```

common/kadai-common/src/test/java/io/kadai/common/internal/util/LambdaCheckedExceptionsTest.java
```diff
@@ -0,0 +1,291 @@
...
+          ClassNotFoundException.class,
+          InterruptedException.class,
+          NoSuchMethodException.class,
+          TimeoutException.class,
+          ConcurrencyException.class,
+          DomainNotFoundException.class
+        })
+    <T extends Exception> void should_RethrowExactException_WhenConsumerThrowsCheckedException(
+        Class<T> clazz) throws Exception {
+      final T expected = clazz.getDeclaredConstructor(String.class).newInstance("Some message");
```
jannikFuellgrafEnvite, line 41: You could consider outsourcing the default exception creation to a method to reduce redundancy

jannikFuellgrafEnvite (approved):

PR #648, opened 2025-03-04
Closes #634 - Relax handling of query flags

### LINKED ISSUE #634: Relax handling of Query-Flags
A query flag is a boolean query parameter.
Consider `api?myFlag=true`. 
It's short form is `api?myFlag`, because the non-existence of the flag implies `false` - and it's existence `true`.

Currently Kadai handles these overly strict, e.g. in [`UserController::getUsers`](https://github.com/kadai-io/kadai/blob/master/rest/kadai-rest-spring/src/main/java/io/kadai/user/rest/UserController.java#L83):
```java
 if (filterParameter.getCurrentUser() != null && QueryParamsValidator.hasQueryParameterValues(request, "current-user")) {
    throw new InvalidArgumentException("It is prohibited to use the param current-user with values.");
}
```

We want to relax this behavior by allowing values for query flags and just ignoring them.
In fact, we want to:
- ignore any values for a query flag and treat the flag as set to `true` - except for the value of the flag being `"false"`,
- treat `api?myFlag=` and `api?myFlag` as `true` with the value `""`

Check for usage of Query-Flags in all REST-Controllers and adjust accordingly.
Add tests for above conditions.

bruderj15:
## Definition of Done

  - will be squashed on merge


rest/kadai-rest-spring/src/main/java/io/kadai/task/rest/TaskQueryFilterParameter.java
```diff
@@ -817,11 +818,12 @@ public class TaskQueryFilterParameter implements QueryParameter<TaskQuery, Void>
   @JsonProperty("owner-not-like")
   private final String[] ownerNotLike;
 
-  @Schema(
+  @Parameter(
```
jamesrdi, line 821: Now that you mention this, I think we should change all the @Schemas with @Parameter, but maybe in a separate PR

rest/kadai-rest-spring-example-boot/src/main/java/io/kadai/example/boot/OpenApiConfiguration.java
```diff
@@ -57,8 +57,9 @@
                     + "</p>"
                     + "<p>"
                     + "Whenever a parameter is a value-less type (e.g owner-is-null and "
-                    + "current-user) it is expected to be defined without a value, i.e., it should"
-                    + " be specified as ?parameter and not ?parameter= or ?parameter=someValue"
+                    + "current-user) it is expected to be defined without a value "
```
jamesrdi, line 60: What do you think of
 "it is expected to be defined either without a value (?parameter), with the empty value (?parameter=), or with the value true (?parameter=true)"

jamesrdi (approved):

rest/kadai-rest-spring-example-boot/src/main/java/io/kadai/example/boot/OpenApiConfiguration.java
jamesrdi, line 60: What do you think of
"it is expected to be defined either without a value (?parameter), with the empty value (?parameter=), or with the value true (?parameter=true)"

jamesrdi (changes requested):

rest/kadai-rest-spring/src/main/java/io/kadai/task/rest/TaskQueryFilterParameter.java
bruderj15: You probably just pinged some random folks ;)
Yeah we should, I added #711.

rest/kadai-rest-spring-example-boot/src/main/java/io/kadai/example/boot/OpenApiConfiguration.java
bruderj15: Literally the same.
Different wording - same semantics.

jamesrdi (approved):

PR #675, opened 2025-03-10
Closes #606 - Improving querying-performance with physical paging

### LINKED ISSUE #606: Improving querying-performance with physical paging
Currently, whenever we want paged results for queries, we use `RowBounds` from MyBatis.
Consider below example for querying users in [UserQueryImpl](https://github.com/kadai-io/kadai/blob/de7609dda173eb0878f7fc275552d6ae06d3ce42/lib/kadai-core/src/main/java/io/kadai/user/internal/UserQueryImpl.java#L86):
```java
public class UserQueryImpl implements UserQuery {
// ...
  @Override
  public List<User> list(int offset, int limit) {
    try {
      UserServiceImpl userService = (UserServiceImpl) kadaiEngine.getEngine().getUserService();
      kadaiEngine.openConnection();
      RowBounds rowBounds = new RowBounds(offset, limit);
      return kadaiEngine
          .getSqlSession()
          .<UserImpl>selectList(LINK_TO_USER_MAPPER, this, rowBounds)
          .stream()
          .map(
              user -> {
                user.setDomains(userService.determineDomains(user));
                return (User) user;
              })
          .toList();
    } catch (PersistenceException e) {
    // ...
  }
// ...
}
```
This is comfortable.
However, that comfort costs us quite a chunk of performance as the MyBatis-implementation  of `RowBounds` just uses the absolute cursor on Javas `ResultSet` - therefore retrieving all users first and then filtering for the relevant.
This is obviously _logical_ pagination (done by the application - kadai).
No need to mentioned that this is bad for larger datasets, significantly for > 10k entries.

We should consider to make use of _physical_ pagination, where the SQL-Query is structured in such a way that it returns just the relevant entries.
These are database-specific - think of `ROWNUM` in Oracle or `LIMIT/OFFSET` in Postgres.
I am pretty sure we can integrate this with elegance and keep the comfort - probably with interceptors.

There is a big plugin doing exactly that: [pagehelper/Mybatis-PageHelper](https://github.com/pagehelper/Mybatis-PageHelper/blob/master/wikis/en/HowToUse.md)
It supports DB2, Postgres, H2 and some more. 
Their JavaDoc is chinese but all documentation is english.

On the first glance it looks like Plug-n-Play.
They detect the database engine and convert the `RowBounds` to native SQL-Queries.
Therefore I expect only minimal changes to be necessary on our side.

jannikFuellgrafEnvite:
## Definition of Done

  - will be squashed on merge

### DIFF
--- a/lib/kadai-core/src/main/java/io/kadai/common/internal/PaginationInterceptor.java
+++ b/lib/kadai-core/src/main/java/io/kadai/common/internal/PaginationInterceptor.java
@@ -1,0 +1,123 @@
+package io.kadai.common.internal;
+
+import java.sql.Connection;
+import org.apache.ibatis.executor.statement.StatementHandler;
+import org.apache.ibatis.mapping.BoundSql;
+import org.apache.ibatis.mapping.MappedStatement;
+import org.apache.ibatis.plugin.Interceptor;
+import org.apache.ibatis.plugin.Intercepts;
+import org.apache.ibatis.plugin.Invocation;
+import org.apache.ibatis.plugin.Signature;
+import org.apache.ibatis.reflection.MetaObject;
+import org.apache.ibatis.reflection.SystemMetaObject;
+
+@Intercepts({
+  @Signature(
+      type = StatementHandler.class,
+      method = "prepare",
+      args = {Connection.class, Integer.class})
+})
+public class PaginationInterceptor implements Interceptor {
+  private static final ThreadLocal<Pagination> PAGINATION_THREAD_LOCAL = new ThreadLocal<>();
+
+  public static void setPagination(int offset, int limit) {
+    int safeLimit = Math.max(0, limit);
+    int safeOffset = Math.max(0, offset);
+    PAGINATION_THREAD_LOCAL.set(new Pagination(safeOffset, safeLimit));
+  }
+
+  public static void clearPagination() {
+    PAGINATION_THREAD_LOCAL.remove();
+  }
+
+  @Override
+  public Object intercept(Invocation invocation) throws Throwable {
+    if (PAGINATION_THREAD_LOCAL.get() == null) {
+      return invocation.proceed();
+    }
+
+    StatementHandler statementHandler = (StatementHandler) invocation.getTarget();
+    MetaObject metaObject = SystemMetaObject.forObject(statementHandler);
+
+    while (metaObject.hasGetter("delegate")) {
+      statementHandler = (StatementHandler) metaObject.getValue("delegate");
+      metaObject = SystemMetaObject.forObject(statementHandler);
+    }
+
+    BoundSql boundSql = statementHandler.getBoundSql();
+    String originalSql = boundSql.getSql();
+
+    if (originalSql.toLowerCase().contains("limit")
+        || originalSql.toLowerCase().contains("offset")) {
+      return invocation.proceed();
+    }
+
+    if (!metaObject.hasGetter("mappedStatement")) {
+      throw new IllegalStateException("DatabaseId could not be determined.");
+    }
+
+    MappedStatement mappedStatement = (MappedStatement) metaObject.getValue("mappedStatement");
+    String databaseId = mappedStatement.getConfiguration().getDatabaseId();
+
+    Pagination pagination = PAGINATION_THREAD_LOCAL.get();
+    String paginatedSql = getPagninatedString(originalSql, databaseId, pagination);
+
+    BoundSql newBoundSql =
+        new BoundSql(
+            mappedStatement.getConfiguration(),
+            paginatedSql,
+            boundSql.getParameterMappings(),
+            boundSql.getParameterObject());
+
+    metaObject.setValue("boundSql", newBoundSql);
+
+    Object result = invocation.proceed();
+    clearPagination();
+    return result;
+  }
+
+  private static String getPagninatedString(
+      String originalSql, String databaseId, Pagination pagination) {
+    String paginatedSql;
+
+    if (originalSql.toLowerCase().contains("order by")) {
+      paginatedSql =
+          originalSql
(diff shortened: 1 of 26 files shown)


lib/kadai-core/src/main/java/io/kadai/common/internal/PaginationInterceptor.java
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    public int getOffset() {
+      return offset;
+    }
+
+    public int getLimit() {
+      return limit;
+    }
+  }
```
bruderj15, line 122: One could argue two points here:
- keep the class but make it a record
- get rid of the class and just use [`RowBounds`](https://github.com/mybatis/mybatis-3/blob/master/src/main/java/org/apache/ibatis/session/RowBounds.java)

I'd prefer the first and also renaming the record to `PhysicalBounds` to make it's intent clear.

lib/kadai-core/src/main/java/io/kadai/common/internal/PaginationInterceptor.java
```diff
@@ -0,0 +1,123 @@
...
+            mappedStatement.getConfiguration(),
+            paginatedSql,
+            boundSql.getParameterMappings(),
+            boundSql.getParameterObject());
+
+    metaObject.setValue("boundSql", newBoundSql);
+
+    Object result = invocation.proceed();
+    clearPagination();
+    return result;
```
bruderj15, line 76: This makes it look like there's some pagination-side-effect in `invocation.proceed()` and we do need to run it before `clearPagination()`.
What about:
```java
clearPagination();
return invocation.proceed();
```

lib/kadai-core/src/main/java/io/kadai/common/internal/PaginationInterceptor.java
```diff
@@ -0,0 +1,123 @@
...
+    if (originalSql.toLowerCase().contains("order by")) {
+      paginatedSql =
+          originalSql
+              .replaceAll("(?i)ORDER BY\\s+DUE\\s+DESC", "ORDER BY DUE DESC, ID ASC")
+              .replaceAll("(?i)ORDER BY\\s+POR_TYPE\\s+ASC", "ORDER BY POR_TYPE ASC, ID ASC")
+              .replaceAll("(?i)ORDER BY\\s+NAME\\s+ASC", "ORDER BY NAME ASC, ID ASC")
+              .replaceAll("(?i)ORDER BY\\s+PRIORITY\\s+DESC", "ORDER BY PRIORITY DESC, ID ASC")
+              .replaceAll("(?i)ORDER BY\\s+PLANNED\\s+ASC", "ORDER BY PLANNED ASC, ID ASC")
+              .replaceAll("(?i)ORDER BY\\s+RECEIVED\\s+DESC", "ORDER BY RECEIVED DESC, ID ASC")
+              .replaceAll("(?i)ORDER BY\\s+OWNER\\s+ASC", "ORDER BY OWNER ASC, ID ASC");
```
bruderj15, line 92: Why do we *always* need to sort by `ID`?

What if the query already contains `ORDER BY ID DESC`?

lib/kadai-core/src/main/java/io/kadai/common/internal/PaginationInterceptor.java
```diff
@@ -0,0 +1,123 @@
...
+    } else {
+      paginatedSql = originalSql;
+    }
+
+    if ("db2".equalsIgnoreCase(databaseId)) {
+      paginatedSql +=
+          " OFFSET " + pagination.offset + " ROWS FETCH FIRST " + pagination.limit + " ROWS ONLY";
+    } else {
+      paginatedSql += " LIMIT " + pagination.limit + " OFFSET " + pagination.offset;
+    }
```
bruderj15, line 102: Consider using a combination of Factory- and Strategy-Pattern here.
Something like:
```java
interface NativeSQLBoundsTransformer {
  String withNativeBounds(Pagination pagination, String sql);
}
public class OffsetLimitBoundsTransformer implements NativeSQLBoundsTransformer { ... }
public class FetchFirstOnlyBoundsTransformer implements NativeSQLBoundsTransformer { ... }
public class NativeSQLBoundsTransformerFactory {
  private NativeSQLBoundsTransformerFactory() {}
  public static NativeSQLBoundsTransformer valueOf(String databaseId) { ... }
}
```
(extra package)

bruderj15 (commented):
I wonder how much the use of Reflection affects performance here.
It's used quite a lot in `PaginationInterceptor::intercept`.
We shall not forget that this is run on almost every query then.
May it be possible to limit the amount of Reflection used?

The current approach internally rewrites the `BoundsSql` - the most obvious approach - certainly works.
Have you considered intercepting `Executor::query` instead?

lib/kadai-core/src/main/java/io/kadai/common/internal/PaginationInterceptor.java
bruderj15: Whew.
I am running into errors [here](https://github.com/bruderj15/kadai-fork/actions/runs/14046575278/job/39328696890#step:7:3350).
Probably why you order by ID explicitly.

bruderj15: I understand it now.
As we know, when using `RowBounds`, MyBatis gets all entries and then drops the irrelevant.
Since we get all entries first, their order is determined by the queries `ORDER BY` paired with databases internal structure - for most relational databases simply `BY ID ASC` - since there is no partitioning.
This is deterministic.
The MyBatis-implementation for `RowBounds` is deterministic as well - so the whole result is.

However, when we use native sql we introduce non-determinism because the native part (e.g. OFFSET/LIMIT) has undefined behavior from the DBMS-side.
Therefore, explicitly ordering by `ORDER BY ID ASC` solves the problem.

PR #677, opened 2025-03-10
Closes #649 - Make `UserController::getUsers` pageable & sortable

### LINKED ISSUE #649: Make `UserController::getUsers` pageable & sortable
With #459 we introduced querying for users.
Analogous to `TaskController::getTasks` and the other controllers we should add support for paging and sorting.

Probably also needs a `UserSummary`.

leantios:
## Definition of Done

  - will be squashed on merge


lib/kadai-core/src/main/java/io/kadai/user/internal/models/UserImpl.java
```diff
@@ -227,24 +36,23 @@ public UserImpl copy() {
...
-        longName,
-        email,
-        phone,
-        mobilePhone,
-        orgLevel4,
-        orgLevel3,
-        orgLevel2,
-        orgLevel1,
-        data,
-        domains);
```
bruderj15: This dropped `UserImpl::hashCode` entirely and relies on the super?
This may work so far because the super does the exact same.
However that may change in the future and one would probably forget to adjust for proper hashing.

I'd suggest to override it here, simply calling the super.
That way whenever `UserImpl` may add additional attributes, one would stumble upon the explicit super call.

lib/kadai-core/src/test/java/acceptance/user/query/QueryUsersAccTest.java
```diff
@@ -194,12 +202,117 @@ void should_ReturnExistingUsers_For_CombinationOfsUserIdsAndOrgLevelWithEmptyCut
...
-        .orgLevel1In("Bat")
-        .list();
+    List<UserSummary> users =
+        kadaiEngine.getUserService().createUserQuery().idIn(userIds).orgLevel1In("Bat").list();
 
     assertThat(users).isEmpty();
   }
+
+  @ParameterizedTest
+  @CsvSource(
```
bruderj15: JUnit has an elegant source provider for enums:
```java
@ParameterizedTest
@EnumSource(SortDirection.class)
...
```

lib/kadai-core/src/test/java/acceptance/user/query/QueryUsersAccTest.java
```diff
@@ -194,12 +202,117 @@ void should_ReturnExistingUsers_For_CombinationOfsUserIdsAndOrgLevelWithEmptyCut
...
+      delimiter = ';',
+      value = {"ASCENDING", "DESCENDING"})
+  void should_SortUsersByFirstName(SortDirection sortDirection) {
```
bruderj15: Just a picky note: 
I personally like to suffix parametized tests with `_For_...`:
 `should_SortUserByFirstName_For_SortDirection`
 
 This is personal taste and you certainly do not have to share it!

lib/kadai-core/src/test/java/acceptance/user/query/QueryUsersAccTest.java
```diff
@@ -16,22 +18,23 @@
 import org.junit.jupiter.params.provider.CsvSource;
 
 @ExtendWith(JaasExtension.class)
-public class QueryUsersAccTest extends AbstractAccTest {
+class QueryUsersAccTest extends AbstractAccTest {
```
bruderj15, line 37: May add an test class for sorting.
Task has `QueryTasksAccTest`, `QueryTasksWithSortingAccTest`, ...

lib/kadai-core/src/test/java/acceptance/user/query/QueryUsersWithPaginationAccTest.java
```diff
@@ -0,0 +1,139 @@
...
+    assertThat(results).hasSize(3);
+  }
+
+  @Test
+  void testPaginationWithPages() {
+    UserService userService = kadaiEngine.getUserService();
+
+    // Getting full page
+    int pageNumber = 2;
+    int pageSize = 4;
```
bruderj15: I don't think these have to be declared explicitly, you may just inline them.

rest/kadai-rest-spring/src/main/java/io/kadai/user/rest/UserApi.java
```diff
@@ -90,16 +115,20 @@ ResponseEntity<UserRepresentationModel> getUser(@PathVariable("userId") String u
             content =
                 @Content(
                     mediaType = MediaTypes.HAL_JSON_VALUE,
-                    schema = @Schema(implementation = UserCollectionRepresentationModel.class))),
+                    schema =
+                        @Schema(implementation = UserSummaryCollectionRepresentationModel.class))),
```
bruderj15: This should probably be `UserSummaryPagedRepresentationModel`.

rest/kadai-rest-spring/src/test/java/io/kadai/user/rest/UserControllerIntTest.java
```diff
@@ -478,6 +484,90 @@ void should_ReturnIntersection_When_OrgLevelAndUserIdsAndCurrentUserAreGiven(
...
+        Map.of(
+            "FIRST_NAME", UserSummaryRepresentationModel::getFirstName,
+            "LAST_NAME", UserSummaryRepresentationModel::getLastName,
+            "ORG_LEVEL_1", UserSummaryRepresentationModel::getOrgLevel1,
+            "ORG_LEVEL_2", UserSummaryRepresentationModel::getOrgLevel2,
+            "ORG_LEVEL_3", UserSummaryRepresentationModel::getOrgLevel3,
+            "ORG_LEVEL_4", UserSummaryRepresentationModel::getOrgLevel4);
+
+    return sortByMap.get(sortBy);
+  }
```
bruderj15: Always a fan of abstract testing!
Instead of defining the awkward `getUserRepresentationModelStringFunction`, maybe play around with `@MethodSource` and supply the getter as `Argument`.

bruderj15 (commented):
This looks really solid already!
I attached a few minor (mainly cosmetic) notes.

@gitgoodjhe I wonder if `UserSummary` should exclude `data` (additional user information)...?

lib/kadai-core/src/main/java/io/kadai/user/internal/models/UserImpl.java
leantios: Good idea!

rest/kadai-rest-spring/src/test/java/io/kadai/user/rest/UserControllerIntTest.java
leantios: mhh 
`Arguments.of("FIRST_NAME", SortDirection.ASCENDING, 0, 5, 5, (Function<UserSummaryRepresentationModel, String>) UserSummaryRepresentationModel::getFirstName)` does not fit in one line, so the code becomes very long with the correct formatting.

gitgoodjhe:
> This looks really solid already! I attached a few minor (mainly cosmetic) notes.
> 
> @gitgoodjhe I wonder if `UserSummary` should exclude `data` (additional user information)...?

good question and debatable. My first thought was probably not , because it is only a String. However, on the database it is in fact defined as clob...so unlimited space. We usually remove clobs/blobs from summaries. I would like to confirm what usually is the contents of this string before deciding

leantios:
> > This looks really solid already! I attached a few minor (mainly cosmetic) notes.
> > @gitgoodjhe I wonder if `UserSummary` should exclude `data` (additional user information)...?
> 
> good question and debatable. My first thought was probably not , because it is only a String. However, on the database it is in fact defined as clob...so unlimited space. We usually remove clobs/blobs from summaries. I would like to confirm what usually is the contents of this string before deciding

We decided to remove `data` from the `UserSummary`

lib/kadai-core/src/main/java/io/kadai/user/internal/models/UserImpl.java
```diff
@@ -19,213 +19,34 @@
...
+  public UserImpl copy() {
+    return new UserImpl(this);
   }
 
   @Override
   public String getData() {
     return data;
   }
 
-  @Override
```
bruderj15: Probably didn't mean to remove the annot.
I would keep it.

rest/kadai-rest-spring/src/main/java/io/kadai/user/rest/UserController.java
```diff
@@ -124,4 +152,42 @@ public ResponseEntity<UserRepresentationModel> deleteUser(@PathVariable("userId"
...
+    public UserQuerySortParameter(List<UserQuerySortBy> sortBy, List<SortDirection> order)
+        throws InvalidArgumentException {
+      super(sortBy, order);
+    }
+
+    // this getter is necessary for the documentation!
+    @Override
+    public List<UserQuerySortBy> getSortBy() {
+      return super.getSortBy();
+    }
```
bruderj15: You removed this with #711 iirc.
Might also drop it here.

bruderj15 (changes requested):
Cosmetics and we're good to go!

bruderj15 (approved):

PR #686, opened 2025-03-12
Closes #682: Fix Swagger UI Dokumentation for getWorkbaskets and getU…

### LINKED ISSUE #682: Fix Swagger UI Dokumentation for `getWorkbaskets` and `getUsers`
In the [Swagger UI](https://kadai-io.azurewebsites.net/kadai/swagger-ui/index.html), `getWorkbaskets` and `getUsers` (sorting will be available with https://github.com/kadai-io/kadai/issues/649) have the following documentation for the `sort-by` query parameter:
```
sort-by | Sort  the result by a given field. Multiple sort values can be declared. When the primary sort value is the same, the second one will be used.

Available values  : CLASSIFICATION_KEY, CLASSIFICATION_NAME, POR_TYPE, POR_VALUE,  POR_COMPANY, POR_SYSTEM, POR_SYSTEM_INSTANCE, STATE, NAME, DUE, PLANNED,  RECEIVED, PRIORITY, CREATED, CLAIMED, DOMAIN, TASK_ID, MODIFIED,  CREATOR, NOTE, OWNER, OWNER_LONG_NAME, BUSINESS_PROCESS_ID,  PARENT_BUSINESS_PROCESS_ID, WORKBASKET_KEY, CUSTOM_1, CUSTOM_2,  CUSTOM_3, CUSTOM_4, CUSTOM_5, CUSTOM_6, CUSTOM_7, CUSTOM_8, CUSTOM_9,  CUSTOM_10, CUSTOM_11, CUSTOM_12, CUSTOM_13, CUSTOM_14, CUSTOM_15,  CUSTOM_16, WORKBASKET_ID, WORKBASKET_NAME,  ATTACHMENT_CLASSIFICATION_KEY, ATTACHMENT_CLASSIFICATION_NAME,  ATTACHMENT_CLASSIFICATION_ID, ATTACHMENT_CHANNEL, ATTACHMENT_REFERENCE, ATTACHMENT_RECEIVED, COMPLETED
```

The `Available values` are the ones for the task (see `io.kadai.task.rest.TaskController.TaskQuerySortBy`). 

Please update the code so that the correct `Available values` are shown.

jamesrdi:
## Definition of Done

  - will be squashed on merge

### DIFF
--- a/rest/kadai-rest-spring/src/main/java/io/kadai/common/rest/QuerySortParameter.java
+++ b/rest/kadai-rest-spring/src/main/java/io/kadai/common/rest/QuerySortParameter.java
@@ -22,13 +22,15 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 import io.kadai.common.api.BaseQuery;
 import io.kadai.common.api.BaseQuery.SortDirection;
 import io.kadai.common.api.exceptions.InvalidArgumentException;
+import io.swagger.v3.oas.annotations.Parameter;
 import io.swagger.v3.oas.annotations.media.Schema;
 import java.util.List;
 
 public class QuerySortParameter<Q extends BaseQuery<?, ?>, S extends QuerySortBy<Q>>
     implements QueryParameter<Q, Void> {
 
-  @Schema(
+
+  @Parameter(
       name = "sort-by",
       description =
           "Sort the result by a given field. Multiple sort values can be declared. When the "


rest/kadai-rest-spring/src/main/java/io/kadai/classification/rest/ClassificationController.java
```diff
@@ -205,6 +207,31 @@ public ClassificationQuerySortParameter(
...
+                  "CUSTOM_3",
+                  "CUSTOM_4",
+                  "CUSTOM_5",
+                  "CUSTOM_6",
+                  "CUSTOM_7",
+                  "CUSTOM_8",
+                  "NAME",
+                  "PARENT_ID",
+                  "PARENT_KEY",
+                  "PRIIORITY",
```
bruderj15: This highlights why this approach ships with a highly likely source of failures.

bruderj15 (commented):
This fixes the issue, which is good.
However, I think there may be a more elegant solution - removing the redundancy introduced to maintain the parameter-schemata when changing the actual enum.

What we want is to just put the enums values there:
```java
@Parameter(.
  .., 
  schema = @Schema(allowableValues =
    Arrays.stream(
      ClassificationQuerySortBy.values()).map(Enum::name).toArray(String[]::new)
    )
)
```
This obviously doesn't work because the rval of `allowableValues` must be a constant.

Ideally we could dynamically set the `allowableValues` right on `QuerySortParameter::sortBy`.
I think we already use AspectJ with Spring in `kadai-rest-spring` and we can probably use it here.
We surely can with an extra annotation, best would be without one.

Doing this on a specifc `QuerySortParameter` such as `ClassificationQuerySortParameter` is a little easier because we would not need to reify the generically bound `QuerySortBy`.

Do you want to try this approach for `ClassificationQuerySortParameter`?
If this is too abstract I could demonstrate POC or do it completely if you want :)

bruderj15:
Seems like there's a simple fix to this.
Just change `@Schema` to `@Parameter` in `QuerySortParameter`:
```java
public class QuerySortParameter<Q extends BaseQuery<?, ?>, S extends QuerySortBy<Q>>
    implements QueryParameter<Q, Void> {

  @Parameter(
      name = "sort-by",
      description =
          "Sort the result by a given field. Multiple sort values can be declared. When the "
              + "primary sort value is the same, the second one will be used.")
  @JsonProperty("sort-by")
  private final List<S> sortBy;
//...
}
```
I tried it locally and can confirm this works.
(No other adjustments needed at all - compared to master)

jamesrdi:
> Seems like there's a simple fix to this. Just change `@Schema` to `@Parameter` in `QuerySortParameter`:
> 
> ```java
> public class QuerySortParameter<Q extends BaseQuery<?, ?>, S extends QuerySortBy<Q>>
>     implements QueryParameter<Q, Void> {
> 
>   @Parameter(
>       name = "sort-by",
>       description =
>           "Sort the result by a given field. Multiple sort values can be declared. When the "
>               + "primary sort value is the same, the second one will be used.")
>   @JsonProperty("sort-by")
>   private final List<S> sortBy;
> //...
> }
> ```
> 
> I tried it locally and can confirm this works. (No other adjustments needed at all - compared to master)

Yes, you're right, thank you for finding this out :)

bruderj15 (dismissed):

bruderj15:
Reminder from the weekly:
- Remove the old comments regarding auto-rest-docs
- Try removing the explicit declaration of the (previously?) necessary class/methods (`[Task/Workbasket/...]QuerySortParameter`).

jamesrdi:
> Reminder from the weekly:
> 
>     * Remove the old comments regarding auto-rest-docs
> 
>     * Try removing the explicit declaration of the (previously?) necessary class/methods (`[Task/Workbasket/...]QuerySortParameter`).

I'll look at it on Monday, if we can remove the explicit declaration

bruderj15 (approved):

PR #746, opened 2025-03-26
Closes #606 - Improving querying-performance with physical paging

### LINKED ISSUE #606: Improving querying-performance with physical paging

bruderj15:
## Definition of Done

  - will be squashed on merge

### DIFF
--- a/common/kadai-common/pom.xml
+++ b/common/kadai-common/pom.xml
@@ -57,6 +57,11 @@
             <version>${version.assertj}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+          <groupId>org.mockito</groupId>
+          <artifactId>mockito-core</artifactId>
+          <scope>test</scope>
+        </dependency>
     </dependencies>
 
 </project>
--- a/common/kadai-common/src/main/java/io/kadai/common/internal/configuration/DB.java
+++ b/common/kadai-common/src/main/java/io/kadai/common/internal/configuration/DB.java
@@ -54,19 +54,19 @@ public enum DB {
   public final String dbProductId;
   public final String detectionScript;
   public final String schemaScript;
-  public final PageDialect dialect;
+  public final PageDialect pageDialect;
 
   DB(
       String dbProductName,
       String dbProductId,
       String detectionScript,
       String schemaScript,
-      PageDialect dialect) {
+      PageDialect pageDialect) {
     this.dbProductName = dbProductName;
     this.dbProductId = dbProductId;
     this.detectionScript = detectionScript;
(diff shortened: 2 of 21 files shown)


jannikFuellgrafEnvite (changes requested):
Overall, it looks pretty good. The performance tests also show that the solution is even better. However, you still have to consider the case that someone enters negative values for the pagination!

jannikFuellgrafEnvite (approved):

PR #769, opened 2025-03-31
 Closes #739: Add Copyright Headers 

### LINKED ISSUE #739: Automatically adding and updating Copyright-header
Manually adding the copyright-header for source files is tiring.
Manually updating it's year is even worse.

I propose to automate this in CI.
There's probably some Github-Action around already.
If not, just create one.

jamesrdi:
## Definition of Done

  - will be squashed on merge

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -325,6 +325,47 @@ jobs:
         if: failure()
         uses: andymckay/cancel-action@0.5
 
+  add-copyright:
+    runs-on: ubuntu-22.04
+    if: github.repository == 'kadai-io/kadai' && github.ref == 'refs/heads/master' && github.head_ref == ''
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Add Copyright Headers
+        run: |
+          YEAR=$(date +%Y)
+          LAST_YEAR=$((YEAR - 1))
+          for file in $(find . -type f \( -name '*.java' -o -name '*.ts' \)); do
+            if grep -q "Copyright \[$LAST_YEAR\] \[envite consulting GmbH\]" "$file"; then
+              echo "Updating copyright year in $file"
+              sed -i "s/Copyright \[$LAST_YEAR\] \[envite consulting GmbH\]/Copyright [$YEAR] [envite consulting GmbH]/" "$file"
+            elif ! grep -q -e "Copyright" -e "License" "$file"; then
+              echo "Adding copyright to $file"
+              cat LICENSE_HEADER.txt "$file" > temp_file
+              mv temp_file "$file"
+            fi
+          done
+          
+          for file in $(find . -type f -name '*.html'); do
+            if grep -q "Copyright \[$LAST_YEAR\] \[envite consulting GmbH\]" "$file"; then
+              echo "Updating copyright year in $file"
+              sed -i "s/Copyright \[$LAST_YEAR\] \[envite consulting GmbH\]/Copyright [$YEAR] [envite consulting GmbH]/" "$file"
+            elif ! grep -q -e "Copyright" -e "License" "$file"; then
+              echo "Adding copyright to $file"
+              cat LICENSE_HEADER_HTML.txt "$file" > temp_file
+              mv temp_file "$file"
+            fi
+          done
+
+      - name: Commit and push changes
+        run: |
+          git config --global user.name "github-actions[bot]"
+          git config --global user.email "github-actions[bot]@users.noreply.github.com"
+          git add .
+          git commit -m "Add/update copyright headers" || echo "No changes to commit"
+          git push
+
   release_artifacts:
     runs-on: ubuntu-22.04
     name: Release artifacts to Sonatype-Central
--- a/LICENSE_HEADER.txt
+++ b/LICENSE_HEADER.txt
@@ -1,0 +1,18 @@
+/*
+ * Copyright [2025] [envite consulting GmbH]
+ *
+ *    Licensed under the Apache License, Version 2.0 (the "License");
+ *    you may not use this file except in compliance with the License.
+ *    You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ *
+ *
+ */
+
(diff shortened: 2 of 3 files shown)


LICENSE_HEADER.txt
```diff
@@ -0,0 +1,18 @@
+/*
+ * Copyright [2024] [envite consulting GmbH]
```
bruderj15: Should use 2025 for each header

bruderj15 (changes requested):

LICENSE_HEADER.txt
bruderj15: On first sight of the script I guess you would need to remove the old (2024) header manually as the bash-logic just prepends?
Ideally we want to replace the header if it's not what's specified in the `LICENSE_HEADER.*`

.github/workflows/continuous-integration.yml
```diff
@@ -325,6 +325,38 @@ jobs:
         if: failure()
         uses: andymckay/cancel-action@0.5
 
+  add-copyright:
+    runs-on: ubuntu-22.04
+    if: github.repository != 'kadai-io/kadai'
```
CRoberto1926: Why != ?

jamesrdi: Because the github actions cannot directly commit and push to the upstream since it is unauthorized, and it will make the github pipeline in the upstream fail, it can only commit and push to the fork

LICENSE_HEADER.txt
jamesrdi: But there are some classes that are copied from other projects and have their own license, e.g see XslxWorksheetConverter.java. And we don't want to replace the license with our own license, right?

bruderj15: Right.
This is the point where a custom github action would shine.
One could easily configure such logic or supply files/dirs to ignore.

.github/workflows/continuous-integration.yml
```diff
@@ -325,6 +325,49 @@ jobs:
...
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Add Copyright Headers
+        run: |
+          for file in $(find . -type f \( -name '*.java' -o -name '*.ts' \)); do
+            if grep -q "Copyright \[2024\] \[envite consulting GmbH\]" "$file"; then
```
bruderj15: Can you capture the year with a regex so we don't have to change the logic every year?

jamesrdi: Ok I'll try

jamesrdi: I have changed it, please recheck

bruderj15: Yes! This does the job.
However, the regex matches every year, so we search and replace even if the year is correct already.
This just doesn't result in a change so we don't notice - but still does the `sed` for all files that have our copyright already.

You already get `YEAR` env var.
What about matching `YEAR - 1` and then replacing?
This would be invalid if there's no PR for a whole year - well then it's over anyways.
What do you think?

jamesrdi: You're right, good idea, I have changed it

.github/workflows/continuous-integration.yml
```diff
@@ -325,6 +325,47 @@ jobs:
...
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Add Copyright Headers
+        run: |
+          YEAR=$(date +%Y)
+          LAST_YEAR=$((YEAR - 1))
+          for file in $(find . -type f \( -name '*.java' -o -name '*.ts' \)); do
+            if grep -q "Copyright \[[0-9]\{4\}\] \[envite consulting GmbH\]" "$file"; then
```
bruderj15: We can match `LAST_YEAR` here too?
So we don't echo "Updating ..." when already correct.

CRoberto1926:
Just my 2 cents here: why can't we just use our IDEs and let them adjust the copyright headers if we touch a file? This is totally possible and configurable, at least in IntelliJ IDEA.

Or there are any legal bindings, that we have to keep each copyright header in each file properly updated?

bruderj15:
> Just my 2 cents here: why can't we just use our IDEs and let them adjust the copyright headers if we touch a file? This is totally possible and configurable, at least in IntelliJ IDEA.
> 
> Or there are any legal bindings, that we have to keep each copyright header in each file properly updated?

There aren't.
I think we either do it clean and update every file automatically - like this PR does - or we get rid of them entirely.

CRoberto1926:
> > Just my 2 cents here: why can't we just use our IDEs and let them adjust the copyright headers if we touch a file? This is totally possible and configurable, at least in IntelliJ IDEA.
> 
> > 
> 
> > Or there are any legal bindings, that we have to keep each copyright header in each file properly updated?
> 
> 
> 
> There aren't.
> 
> I think we either do it clean and update every file automatically - like this PR does - or we get rid of them entirely.

We can't get rid of them due to licensing issues because of the TASKANA fork

bruderj15:
> We can't get rid of them due to licensing issues because of the TASKANA fork

Well, then I highly endorse this PR.
Takes the load off manual and local configuration, guarantees it for all contributions and keeps license up-to-date.
However @jamesrdi I would only want this to run on push to master.

.github/workflows/continuous-integration.yml
```diff
@@ -325,6 +325,47 @@ jobs:
...
+          LAST_YEAR=$((YEAR - 1))
+          for file in $(find . -type f \( -name '*.java' -o -name '*.ts' \)); do
+            if grep -q "Copyright \[$LAST_YEAR\] \[envite consulting GmbH\]" "$file"; then
+              echo "Updating copyright year in $file"
+              sed -i "s/Copyright \[$LAST_YEAR\] \[envite consulting GmbH\]/Copyright [$YEAR] [envite consulting GmbH]/" "$file"
+            elif ! grep -q -e "Copyright" -e "License" "$file"; then
+              echo "Adding copyright to $file"
+              cat LICENSE_HEADER.txt "$file" > temp_file
+              mv temp_file "$file"
+            fi
```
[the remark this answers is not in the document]
bruderj15: This is fine as we process sequentially

.github/workflows/continuous-integration.yml
```diff
@@ -325,6 +325,47 @@ jobs:
...
+          
+          for file in $(find . -type f -name '*.html'); do
+            if grep -q "Copyright \[$LAST_YEAR\] \[envite consulting GmbH\]" "$file"; then
+              echo "Updating copyright year in $file"
+              sed -i "s/Copyright \[$LAST_YEAR\] \[envite consulting GmbH\]/Copyright [$YEAR] [envite consulting GmbH]/" "$file"
+            elif ! grep -q -e "Copyright" -e "License" "$file"; then
+              echo "Adding copyright to $file"
+              cat LICENSE_HEADER_HTML.txt "$file" > temp_file
+              mv temp_file "$file"
+            fi
```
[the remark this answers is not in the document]
bruderj15: This is fine as we process sequentially

bruderj15 (approved):
I like this.
Let's get it working!

PR #772, opened 2025-03-31
Closes #771 - Reduce Cognitive Complexity in `TaskQueryFilterParameter::validateFilterParameters`

### LINKED ISSUE #771: Reduce Cognitive Complexity in `TaskQueryFilterParameter::validateFilterParameters`
The method `validateFilterParameters` has a cognitive complexity of 81.
 This can be easily handled by using extra methods
The same is true for the method `TaskQueryFilterCustomIntFields::validateFilterParameters`

leantios:
## Definition of Done

  - will be squashed on merge


bruderj15 (commented):
I highly endorse these changes.

What about removing the validation-logic from the parameter?
The current usage of any `QueryFilterParameter` is data-oriented - a record-like class with metadata (for docs).

What do you think about an additional class handling the validation, e.g.:
```java
interface QueryParameterValidation<E, R, Q extends QueryParameter<E, R>> {
  void validate(Q query) throws InvalidArgumentException;
}
```

This would separate the logic and also align well with other (or future) Query-Validations, e.g. [`TaskHistoryQueryFilterParameter`](https://github.com/kadai-io/kadai/blob/master/history/kadai-simplehistory-rest-spring/src/main/java/io/kadai/simplehistory/rest/TaskHistoryQueryFilterParameter.java#L657).

leantios:
> I highly endorse these changes.
> 
> What about removing the validation-logic from the parameter? The current usage of any `QueryFilterParameter` is data-oriented - a record-like class with metadata (for docs).
> 
> What do you think about an additional class handling the validation, e.g.:
> 
> ```java
> interface QueryParameterValidation<E, R, Q extends QueryParameter<E, R>> {
>   void validate(Q query) throws InvalidArgumentException;
> }
> ```
> 
> This would separate the logic and also align well with other (or future) Query-Validations, e.g. [`TaskHistoryQueryFilterParameter`](https://github.com/kadai-io/kadai/blob/master/history/kadai-simplehistory-rest-spring/src/main/java/io/kadai/simplehistory/rest/TaskHistoryQueryFilterParameter.java#L657).

@gitgoodjhe what do you think, would this be a good idea?

gitgoodjhe:
> > I highly endorse these changes.
> > What about removing the validation-logic from the parameter? The current usage of any `QueryFilterParameter` is data-oriented - a record-like class with metadata (for docs).
> > What do you think about an additional class handling the validation, e.g.:
> > ```java
> > interface QueryParameterValidation<E, R, Q extends QueryParameter<E, R>> {
> >   void validate(Q query) throws InvalidArgumentException;
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This would separate the logic and also align well with other (or future) Query-Validations, e.g. [`TaskHistoryQueryFilterParameter`](https://github.com/kadai-io/kadai/blob/master/history/kadai-simplehistory-rest-spring/src/main/java/io/kadai/simplehistory/rest/TaskHistoryQueryFilterParameter.java#L657).
> 
> @gitgoodjhe what do you think, would this be a good idea?

I really like these changes, good job! A separate class is also a neat idea

rest/kadai-rest-spring/src/main/java/io/kadai/task/rest/TaskQueryFilterCustomIntFields.java
```diff
@@ -656,7 +653,7 @@ public TaskQueryFilterCustomIntFields(
     this.customInt8To = customInt8To;
     this.customInt8ToNot = customInt8ToNot;
 
-    validateFilterParameters();
+    new TaskQueryFilterCustomIntFieldsValidation().validate(this);
```
bruderj15: Each implementation of `QueryParameterValidation` is supposed to be pure - without side-effects.
Therefore it shall be enough to allocate memory only once and for only one `TaskQueryFilterCustomIntFieldsValidation` - reusing it.
It should be a private static member of `TaskQueryFilterCustomIntFields`.

leantios: Tests fail -
```
Error:  Failures: 
Error:    SpringArchitectureTest.should_AnnotateAllFieldsWithJsonProperty_When_ImplementingQueryParameter:53 Architecture Violation [Priority: MEDIUM] - Rule 'classes that implement io.kadai.common.rest.QueryParameter should all fields should have a @JsonProperty or @JsonIgnore annotation' was violated (2 times):
Field 'JavaField{io.kadai.task.rest.TaskQueryFilterCustomIntFields.VALIDATOR}' in class 'JavaClass{name='io.kadai.task.rest.TaskQueryFilterCustomIntFields'}' is not annotated with a json annotation
Field 'JavaField{io.kadai.task.rest.TaskQueryFilterParameter.VALIDATOR}' in class 'JavaClass{name='io.kadai.task.rest.TaskQueryFilterParameter'}' is not annotated with a json annotation
```

leantios: When using the interface, making the method static is also not possible. There does not seem to be a nice solution.

bruderj15: Seems like one should adjust the Arch-Rule.
The rule is overly strict.
It should not consider static fields - as they're never mapped - hence can't even be a JSON-Field.
Can you change that?

bruderj15 (approved):
Great job!

PR #793, opened 2025-04-01
Closes #768 - Use Constants for `ColumnName`s

### LINKED ISSUE #768: Use constants for Column Names
In `TaskCommentQuerySqlProvider` and `TaskQuerySqlProvider` Strings are used that already exist as constants in `TaskCommentQueryColumnName` and `TaskQueryColumnName`.
E.g. `tc.ID` or `t.PRIORITY`

leantios:
## Definition of Done

  - will be squashed on merge

### DIFF
--- a/common/kadai-common/src/main/java/io/kadai/common/internal/util/SqlProviderUtil.java
+++ b/common/kadai-common/src/main/java/io/kadai/common/internal/util/SqlProviderUtil.java
@@ -42,7 +42,7 @@ public class SqlProviderUtil {
         .append("</when>")
         .append("<otherwise>0=1</otherwise>")
         .append("</choose>");
-    if (column.matches("t.CUSTOM_\\d+") || column.matches("t.OWNER")) {
+    if (column.matches("t.(?i)custom(?-i)_\\d+") || column.matches("t.(?i)owner(?-i)")) {
       sb.append("<if test='" + collection + "ContainsNull'> OR " + column + " IS NULL </if>");
     }
     return sb.append(")</if> ");
@@ -65,7 +65,7 @@ public class SqlProviderUtil {
         .append("</when>")
         .append("<otherwise>1=1</otherwise>")
         .append("</choose>");
-    if (column.matches("t.CUSTOM_\\d+")) {
+    if (column.matches("t.(?i)custom(?-i)_\\d+")) {
       sb.append("<if test='" + collection + "ContainsNull'> AND " + column + " IS NOT NULL </if>");
       sb.append("<if test='!" + collection + "ContainsNull'> OR " + column + " IS NULL </if>");
     }
--- a/lib/kadai-core/src/main/java/io/kadai/task/internal/ObjectReferenceQuerySqlProvider.java
+++ b/lib/kadai-core/src/main/java/io/kadai/task/internal/ObjectReferenceQuerySqlProvider.java
@@ -24,10 +24,6 @@ import static io.kadai.common.internal.util.SqlProviderUtil.DB2_WITH_UR;
 import static io.kadai.common.internal.util.SqlProviderUtil.OPENING_SCRIPT_TAG;
 import static io.kadai.common.internal.util.SqlProviderUtil.OPENING_WHERE_TAG;
 import static io.kadai.common.internal.util.SqlProviderUtil.whereIn;
-import static io.kadai.task.api.ObjectReferenceQueryColumnName.COMPANY;
-import static io.kadai.task.api.ObjectReferenceQueryColumnName.COMPANY;
-import static io.kadai.task.api.ObjectReferenceQueryColumnName.COMPANY;
-import static io.kadai.task.api.ObjectReferenceQueryColumnName.COMPANY;
 
 
 
--- a/lib/kadai-core/src/main/java/io/kadai/task/internal/ObjectReferenceQuerySqlProvider.java
+++ b/lib/kadai-core/src/main/java/io/kadai/task/internal/ObjectReferenceQuerySqlProvider.java
@@ -24,6 +24,12 @@ import static io.kadai.common.internal.util.SqlProviderUtil.DB2_WITH_UR;
 import static io.kadai.common.internal.util.SqlProviderUtil.OPENING_SCRIPT_TAG;
 import static io.kadai.common.internal.util.SqlProviderUtil.OPENING_WHERE_TAG;
 import static io.kadai.common.internal.util.SqlProviderUtil.whereIn;
+import static io.kadai.task.api.ObjectReferenceQueryColumnName.COMPANY;
+import static io.kadai.task.api.ObjectReferenceQueryColumnName.COMPANY;
+import static io.kadai.task.api.ObjectReferenceQueryColumnName.COMPANY;
+import static io.kadai.task.api.ObjectReferenceQueryColumnName.COMPANY;
+
+
 
 public class ObjectReferenceQuerySqlProvider {
 
--- a/lib/kadai-core/src/main/java/io/kadai/task/internal/TaskCommentQuerySqlProvider.java
+++ b/lib/kadai-core/src/main/java/io/kadai/task/internal/TaskCommentQuerySqlProvider.java
@@ -29,6 +29,12 @@ import static io.kadai.common.internal.util.SqlProviderUtil.whereLike;
 import static io.kadai.common.internal.util.SqlProviderUtil.whereNotIn;
 import static io.kadai.common.internal.util.SqlProviderUtil.whereNotInInterval;
 import static io.kadai.common.internal.util.SqlProviderUtil.whereNotLike;
+import static io.kadai.task.api.TaskCommentQueryColumnName.CREATED;
+import static io.kadai.task.api.TaskCommentQueryColumnName.ID;
+import static io.kadai.task.api.TaskCommentQueryColumnName.TASK_ID;
+import static io.kadai.task.api.TaskCommentQueryColumnName.CREATOR;
+import static io.kadai.task.api.TaskCommentQueryColumnName.MODIFIED;
+import static io.kadai.task.api.TaskCommentQueryColumnName.TEXT_FIELD;
 
 import io.kadai.task.api.TaskCommentQueryColumnName;
 import java.util.Arrays;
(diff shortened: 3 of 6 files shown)


leantios:
I am not sure on how thoroughly I should replace the Strings with columnNames. E.g. in `UserMapperSqlProvider`, in the following function:
```  public static String insertGroups() {
    return OPENING_SCRIPT_TAG
        + "INSERT INTO GROUP_INFO (USER_ID, GROUP_ID) VALUES "
        + "<foreach item='group' collection='groups' open='(' separator='),(' close=')'>"
        + "#{id}, #{group}"
        + "</foreach> "
        + CLOSING_SCRIPT_TAG;
  }
```
I don't think that there exists a `ColumnName` class that contains `GROUP_ID`. Should I create one? And should I also replace `USER_ID`?
What's your opinion here @bruderj15 ?

lib/kadai-core/src/main/java/io/kadai/user/internal/UserMapperSqlProvider.java
```diff
@@ -92,12 +103,34 @@ public static String insertPermissions() {
...
+        + " = #{orgLevel3}, "
+        + UserQueryColumnName.ORG_LEVEL_2
+        + " = #{orgLevel2}, "
+        + UserQueryColumnName.ORG_LEVEL_1
+        + " = #{orgLevel1}, "
+        + UserQueryColumnName.DATA
+        + " = #{data} "
+        + "WHERE "
+        + UserQueryColumnName.USER_ID
+        + " = #{id} ";
```
bruderj15: For readability I'd prefer static imports of the enums values where possible.

bruderj15 (commented):
> I am not sure on how thoroughly I should replace the Strings with columnNames. E.g. in `UserMapperSqlProvider`, in the following function:
> 
> ```
>     return OPENING_SCRIPT_TAG
>         + "INSERT INTO GROUP_INFO (USER_ID, GROUP_ID) VALUES "
>         + "<foreach item='group' collection='groups' open='(' separator='),(' close=')'>"
>         + "#{id}, #{group}"
>         + "</foreach> "
>         + CLOSING_SCRIPT_TAG;
>   }
> ```
> 
> I don't think that there exists a `ColumnName` class that contains `GROUP_ID`. Should I create one? And should I also replace `USER_ID`? What's your opinion here @bruderj15 ?

Good question.
I'd say if there's multiple usages, another enum should be created.
If there's just one, it seems over the top.

lib/kadai-core/src/main/java/io/kadai/user/internal/UserMapperSqlProvider.java
```diff
@@ -21,12 +21,36 @@
...
-      "USER_ID, FIRST_NAME, LASTNAME, FULL_NAME, LONG_NAME, E_MAIL, PHONE, MOBILE_PHONE, "
-          + "ORG_LEVEL_4, ORG_LEVEL_3, ORG_LEVEL_2, ORG_LEVEL_1, DATA ";
+      Arrays.stream(UserQueryColumnName.values())
+              .filter(
+                  column ->
+                      !EnumSet.of(UserQueryColumnName.GROUPS, UserQueryColumnName.PERMISSIONS)
+                          .contains(column))
+              .map(UserQueryColumnName::toString)
+              .collect(Collectors.joining(", "))
+          + " ";
```
bruderj15, line 53: may use [`Collectors::collectingAndThen`](https://docs.oracle.com/javase/8/docs/api/java/util/stream/Collectors.html#collectingAndThen-java.util.stream.Collector-java.util.function.Function-)

lib/kadai-core/src/main/java/io/kadai/task/internal/ObjectReferenceQuerySqlProvider.java
```diff
@@ -25,6 +25,8 @@
 import static io.kadai.common.internal.util.SqlProviderUtil.OPENING_WHERE_TAG;
 import static io.kadai.common.internal.util.SqlProviderUtil.whereIn;
 
+
+
```
bruderj15, line 29: whoopsie

common/kadai-common/src/main/java/io/kadai/common/internal/util/SqlProviderUtil.java
```diff
@@ -65,7 +65,7 @@ public static StringBuilder whereNotIn(String collection, String column, StringB
         .append("</when>")
         .append("<otherwise>1=1</otherwise>")
         .append("</choose>");
-    if (column.matches("t.CUSTOM_\\d+")) {
+    if (column.matches("t.(?i)custom(?-i)_\\d+")) {
```
bruderj15, line 68: Although technically right, this hinders readability.
Is there an invocation where `column` is lower-case?
This should be `apiSpec` - per convention only supply uppercase (clients responsibility).

lib/kadai-core/src/main/java/io/kadai/task/internal/TaskCommentQuerySqlProvider.java
```diff
@@ -101,21 +107,21 @@ private static String commonSelectFields() {
...
+    whereLike("textFieldLike", TEXT_FIELD.toString(), sb);
+    whereNotLike("textFieldNotLike", TEXT_FIELD.toString(), sb);
+    whereIn("creatorIn", CREATOR.toString(), sb);
+    whereNotIn("creatorNotIn", CREATOR.toString(), sb);
+    whereLike("creatorLike", CREATOR.toString(), sb);
+    whereNotLike("creatorNotLike", CREATOR.toString(), sb);
+    whereInInterval("createdIn", CREATED.toString(), sb);
+    whereNotInInterval("createdNotIn", CREATED.toString(), sb);
+    whereInInterval("modifiedIn", MODIFIED.toString(), sb);
+    whereNotInInterval("modifiedNotIn", MODIFIED.toString(), sb);
```
bruderj15, line 124: Can we get rid of explicit `toString()` here?
Java should implicitly type it as String - does it?

lib/kadai-core/src/main/java/io/kadai/user/internal/UserMapperSqlProvider.java
```diff
@@ -38,7 +62,9 @@ public static String findById() {
         + "SELECT "
         + USER_INFO_COLUMNS
         + "FROM USER_INFO "
-        + "WHERE USER_ID = #{id} "
+        + "WHERE "
+        + USER_ID
+        + " = #{id} "
```
bruderj15, line 67: For readability I'd prefer:
```java
   ...
 + "WHERE "
 + USER_ID + " = #{id} "
 + ...
```

bruderj15: What does our Check-Style say?

bruderj15 (commented):
I like the ease & safety of maintenance this introduces.
I dont quite yet buy the readbility aspect.

I guess the check-style auto-formatted the Query-Strings?
```
        + "FROM (SELECT "
        + ID
        + ", "
        + POR_VALUE
        + " </if> "
```

common/kadai-common/src/main/java/io/kadai/common/internal/util/SqlProviderUtil.java
bruderj15: client = context invocator

leantios:
> I like the ease & safety of maintenance this introduces. I dont quite yet buy the readbility aspect.
> 
> I guess the check-style auto-formatted the Query-Strings?
> 
> ```
>         + "FROM (SELECT "
>         + ID
>         + ", "
>         + POR_VALUE
>         + " </if> "
> ```

I agree.
Yes, it's the auto-formatting. If formatted differently, it still passes checkstyle. However, I don't like keeping a solution that is non-compliant with the auto-fomatting.

lib/kadai-core/src/main/java/io/kadai/user/internal/UserMapperSqlProvider.java
leantios: It passes check-style but not auto-formatting

lib/kadai-core/src/main/java/io/kadai/task/internal/TaskCommentQuerySqlProvider.java
leantios: Nope, that does not work sadly. 
```
java: incompatible types: io.kadai.task.api.TaskQueryColumnName cannot be converted to java.lang.String
```
An (also not so pretty) alternative is `String.valueOf(ID)`

bruderj15: Out of curiosity, I guess
```java
CREATOR + ""
```
works?

bruderj15: You may overload `SQLProviderUtil::whereNotLike/...` though and have it accept an enum value.

bruderj15:
```java
  public static StringBuilder whereIn(String collection, String column, StringBuilder sb) {
    sb.append("<if test='")
        .append(collection)
        .append(" != null'>AND (")
        .append("<choose>")
        .append("<when test='" + collection + ".length > 0'>")
        .append(column)
        .append(" IN(<foreach item='item' collection='")
        .append(collection)
        .append("' separator=',' >#{item}</foreach>)")
        .append("</when>")
        .append("<otherwise>0=1</otherwise>")
        .append("</choose>");
    if (column.matches("t.CUSTOM_\\d+") || column.matches("t.OWNER")) {
      sb.append("<if test='" + collection + "ContainsNull'> OR " + column + " IS NULL </if>");
    }
    return sb.append(")</if> ");
  }

  public static StringBuilder whereIn(String collection, String column) {
    return whereIn(collection, column, new StringBuilder());
  }

  public static StringBuilder whereIn(String collection, Enum<?> column) {
    return whereIn(collection, column.name(), new StringBuilder());
  }
```

bruderj15: consider usage of `Enum::name` or `Enum::toString` - probably the latter here

leantios:
`google-java-format` can't be personalized. Readability is extremely important for SQL, so we won't continue with this issue.

PR #795, opened 2025-04-02
Closes #790 - Update RestController annotations to be correct and consistent

### LINKED ISSUE #790: Update RestController annotations to be correct and consistent

leantios:
## Definition of Done

  - will be squashed on merge

### DIFF
--- a/rest/kadai-rest-spring/src/main/java/io/kadai/task/rest/TaskController.java
+++ b/rest/kadai-rest-spring/src/main/java/io/kadai/task/rest/TaskController.java
@@ -137,7 +137,6 @@ public class TaskController implements TaskApi {
   // region READ
 
   @GetMapping(path = RestEndpoints.URL_TASKS)
-  @Transactional(readOnly = true, rollbackFor = Exception.class)
   public ResponseEntity<TaskSummaryPagedRepresentationModel> getTasks(
       HttpServletRequest request,
       @ParameterObject TaskQueryFilterParameter filterParameter,
@@ -172,7 +171,6 @@ public class TaskController implements TaskApi {
   }
 
   @GetMapping(path = RestEndpoints.URL_TASKS_ID)
-  @Transactional(readOnly = true, rollbackFor = Exception.class)
   public ResponseEntity<TaskRepresentationModel> getTask(@PathVariable("taskId") String taskId)
       throws TaskNotFoundException, NotAuthorizedOnWorkbasketException {
     Task task = taskService.getTask(taskId);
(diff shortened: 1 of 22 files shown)


rest/kadai-rest-spring/src/main/java/io/kadai/task/rest/TaskController.java
```diff
@@ -295,6 +295,7 @@ public ResponseEntity<TaskRepresentationModel> forceRequestReview(
     return ResponseEntity.ok(taskRepresentationModelAssembler.toModel(task));
   }
 
+  @PostMapping(path = RestEndpoints.URL_TASKS_ID_REQUEST_CHANGES)
```
bruderj15, line 296: There's even more missing I just noticed: `@Transactional(rollbackFor = Exception.class)`

bruderj15 (changes requested):
Might also check all other controller endpoints to see if these annotations are missing elsewhere.

rest/kadai-rest-spring/src/main/java/io/kadai/task/rest/TaskController.java
leantios: For `@GetMapping`, should we always explicitely model ` @Transactional(readOnly = true, rollbackFor = Exception.class)
`? This is not consistent across the codebase. E.g. `AccessIdController` vs `ClassificationController`

bruderj15: As per specification the HTTP-GET is safe, see [RFC7231/4.2.1](https://datatracker.ietf.org/doc/html/rfc7231#section-4.2.1):
>  Request methods are considered "safe" if their defined semantics are
   essentially read-only; i.e., the client does not request, and does
   not expect, any state change on the origin server as a result of
   applying a safe method to a target resource.
   ...
   Of the request methods defined by this specification, the GET, HEAD,
   OPTIONS, and TRACE methods are defined to be safe.

Therefore it's not only supposed to be idempotent but also pure.
The right answer would be to remove any `@Transactional` annotation from endpoints offering `GET`.
Well, does Kadai follow this spec? One would need to check.

leantios: It seems like the older classes always use `@Transactional`, whereas the newer classes do not use it for `@GetMapping`. Maybe there used to be a rule?

bruderj15: Probably just a dev error.
I'd remove it.

bruderj15 (approved):

PR #831, opened 2025-04-08
Closes #786 - Simplify `ClassificationDefinitionController::mapChildrenToParentKeys`...

### LINKED ISSUE #786: Simplify `ClassificationDefinitionController::mapChildrenToParentKeys`
One can probably improve readbility and performance utilizing streams in [`ClassificationDefinitionController::mapChildrenToParentKeys`](https://github.com/kadai-io/kadai/blob/e9b89db4473b046740adf70a70ca4f5f1d81b000/rest/kadai-rest-spring/src/main/java/io/kadai/classification/rest/ClassificationDefinitionController.java#L152).

leantios:
## Definition of Done

  - will be squashed on merge

### DIFF
--- a/rest/kadai-rest-spring/src/main/java/io/kadai/classification/rest/ClassificationDefinitionController.java
+++ b/rest/kadai-rest-spring/src/main/java/io/kadai/classification/rest/ClassificationDefinitionController.java
@@ -46,7 +46,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.stream.Collectors;
-import java.util.stream.Stream;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.hateoas.config.EnableHypermediaSupport;
 import org.springframework.http.MediaType;
@@ -153,7 +152,6 @@ public class ClassificationDefinitionController implements ClassificationDefinit
   private Map<Classification, String> mapChildrenToParentKeys(
       Collection<ClassificationDefinitionRepresentationModel> definitionList,
       Map<String, String> systemIds) {
-    Stream<ClassificationDefinitionRepresentationModel> definitionStream = definitionList.stream();
 
     Set<String> keysWithDomain =
         definitionList.stream()
@@ -163,12 +161,17 @@ public class ClassificationDefinitionController implements ClassificationDefinit
 
     Map<String, String> classificationIdToKey =
         definitionList.stream()
+            .filter(
+                def ->
+                    def.getClassification().getClassificationId() != null
+                        && def.getClassification().getKey() != null)
             .collect(
                 Collectors.toMap(
                     def -> def.getClassification().getClassificationId(),
-                    def -> def.getClassification().getKey()));
+                    def -> def.getClassification().getKey(),
+                    (existing, replacement) -> existing));
 
-    definitionStream
+    definitionList.stream()
         .map(ClassificationDefinitionRepresentationModel::getClassification)
         .map(this::normalizeNullValues)
         .filter(cl -> !cl.getParentId().isEmpty() && cl.getParentKey().isEmpty())
--- a/rest/kadai-rest-spring/src/main/java/io/kadai/classification/rest/ClassificationDefinitionController.java
+++ b/rest/kadai-rest-spring/src/main/java/io/kadai/classification/rest/ClassificationDefinitionController.java
@@ -123,7 +123,8 @@ public class ClassificationDefinitionController implements ClassificationDefinit
   private Map<String, String> getSystemIds() {
     return classificationService.createClassificationQuery().list().stream()
         .collect(
-            Collectors.toMap(i -> i.getKey() + "|" + i.getDomain(), ClassificationSummary::getId));
+            Collectors.toMap(
+                i -> logicalId(i.getKey(), i.getDomain()), ClassificationSummary::getId));
   }
 
   private ClassificationDefinitionCollectionRepresentationModel
@@ -143,7 +144,7 @@ public class ClassificationDefinitionController implements ClassificationDefinit
                 cl -> {
                   String key = cl.getKey();
                   String domain = cl.getDomain();
-                  if (!seen.add(key + "|" + domain)) {
+                  if (!seen.add(logicalId(key, domain))) {
                     throw new ClassificationAlreadyExistException(key, domain);
                   }
                 }));
@@ -155,8 +156,7 @@ public class ClassificationDefinitionController implements ClassificationDefinit
 
     Set<String> keysWithDomain =
         definitionList.stream()
-            .map(
-                def -> def.getClassification().getKey() + "|" + def.getClassification().getDomain())
+            .map(def -> logicalId(def.getClassification()))
             .collect(Collectors.toSet());
 
     Map<String, String> classificationIdToKey =
@@ -202,7 +202,7 @@ public class ClassificationDefinitionController implements ClassificationDefinit
       ClassificationRepresentationModel cl,
       Set<String> keysWithDomain,
       Map<String, String> systemIds) {
-    String parentKeyAndDomain = cl.getParentKey() + "|" + cl.getDomain();
+    String parentKeyAndDomain = logicalId(cl);
     return !cl.getParentKey().isEmpty()
         && (keysWithDomain.contains(parentKeyAndDomain)
             || systemIds.containsKey(parentKeyAndDomain));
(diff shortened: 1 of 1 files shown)


leantios:
@bruderj15 Do you have an idea why the rethrowing is not working?

bruderj15:
> @bruderj15 Do you have an idea why the rethrowing is not working?

I have copied only your changes from `checkForDuplicates` on top of master and the test failing in CI succeeds locally.
Therefore some other changes in this PR seem to interfere - not `rethrowing`.

rest/kadai-rest-spring/src/main/java/io/kadai/classification/rest/ClassificationDefinitionController.java
```diff
@@ -136,50 +135,77 @@ private Map<String, String> getSystemIds() {
   private void checkForDuplicates(
```
bruderj15, line 136: `ClassificationAlreadyExistsException` seems off here - and it was before this PR.
If client supplies a collection with multiple classifications with identical logical name (key + | + domain), and we check here that this collection is unique in terms of logical name, then we cannot even know it exists already.
It should be something like: `DuplicateClassificationException`

bruderj15: The endpoint where this is used `ClassificationDefinitionController::importClassifications` itself is a mess.
The schema docs for `ClassificationDefinitionApi::importClassifications` state, that we overwrite existing classifications.
It makes no sense at all that this endpoint is specified to throw a (checked) `ClassificationAlreadyExistsException`.
This has been done because `ClassificationDefinitionController::insertOrUpdateClassificationsWithoutParent` does this:
```java
      String systemId =
          systemIds.get(classificationRepModel.getKey() + "|" + classificationRepModel.getDomain());
      if (systemId != null) {
        updateExistingClassification(newClassification, systemId);
      } else {
        classificationService.createClassification(newClassification);
      }
```
Whew.
It gets worse as we even document in the swagger docs that this endpoint may fail because of existing classification (which it never should):
```java
        @ApiResponse(
            responseCode = "409",
            description = "ENTITY_NOT_UP_TO_DATE, CLASSIFICATION_ALREADY_EXISTS",
            content = {
              @Content(
                  schema =
                      @Schema(
                          anyOf = {
                            ConcurrencyException.class,
                            ClassificationAlreadyExistException.class
                          }))
            }),
```

rest/kadai-rest-spring/src/main/java/io/kadai/classification/rest/ClassificationDefinitionController.java
```diff
@@ -136,50 +135,77 @@ private Map<String, String> getSystemIds() {
...
   private Map<Classification, String> mapChildrenToParentKeys(
       Collection<ClassificationDefinitionRepresentationModel> definitionList,
       Map<String, String> systemIds) {
-    Map<Classification, String> childrenInFile = new HashMap<>();
-    Set<String> newKeysWithDomain = new HashSet<>();
+
+    Set<String> keysWithDomain =
+        definitionList.stream()
+            .map(
+                def -> def.getClassification().getKey() + "|" + def.getClassification().getDomain())
```
bruderj15: I have seen the `key + | + domain` quite often.
[Here](https://github.com/kadai-io/kadai/blob/bfab3d9a664f4cae506057d6c47e630690e7ab10/rest/kadai-rest-spring/src/main/java/io/kadai/workbasket/rest/WorkbasketDefinitionController.java#L232) it is for workbaskets.
Consider if it makes sense to add this to the model as transient getter.

rest/kadai-rest-spring/src/main/java/io/kadai/classification/rest/ClassificationDefinitionController.java
```diff
@@ -136,50 +135,77 @@ private Map<String, String> getSystemIds() {
...
-    }
+    Set<String> seen = new HashSet<>();
+    definitionList.stream()
+        .map(ClassificationDefinitionRepresentationModel::getClassification)
+        .forEach(
+            rethrowing(
+                cl -> {
+                  String key = cl.getKey();
+                  String domain = cl.getDomain();
+                  if (!seen.add(key + "|" + domain)) {
```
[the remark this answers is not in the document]
leantios: as `key + "|" + domain` is only used to check for duplicates, I think it's save here

rest/kadai-rest-spring/src/main/java/io/kadai/classification/rest/ClassificationDefinitionController.java
leantios: Do you think
- Renaming ClassificationAlreadyExistsException to something like DuplicateClassificationException to better reflect its purpose
- Updating the Swagger documentation to remove the mention of ClassificationAlreadyExistsException as a possible failure scenario
will fix your concerns? I suggest to create a separate issue for this.

bruderj15: This to be handled in https://github.com/kadai-io/kadai/issues/855.

rest/kadai-rest-spring/src/main/java/io/kadai/classification/rest/ClassificationDefinitionController.java
leantios: I decided to go for the solution like for the workbaskets, because the `ClassificationDefinitionController` is the only place where we need it

rest/kadai-rest-spring/src/main/java/io/kadai/classification/rest/ClassificationDefinitionController.java
```diff
@@ -136,50 +136,76 @@ private Map<String, String> getSystemIds() {
...
   private Map<Classification, String> mapChildrenToParentKeys(
       Collection<ClassificationDefinitionRepresentationModel> definitionList,
       Map<String, String> systemIds) {
-    Map<Classification, String> childrenInFile = new HashMap<>();
-    Set<String> newKeysWithDomain = new HashSet<>();
+
+    Set<String> keysWithDomain =
+        definitionList.stream()
+            .map(def -> logicalId(def.getClassification()))
+            .collect(Collectors.toSet());
```
bruderj15: Personal taste: Make it fully functional-style:
```java
definitionList.stream()
    .map(ClassificationDefinitionRepresentationModel::getClassification)
    .map(this::logicalId)
    .collect(toSet());
```
Also for readbility I prefer static imports for all `Collectors::fun`.

rest/kadai-rest-spring/src/main/java/io/kadai/classification/rest/ClassificationDefinitionController.java
```diff
@@ -136,50 +136,76 @@ private Map<String, String> getSystemIds() {
...
+        definitionList.stream()
+            .filter(
+                def ->
+                    def.getClassification().getClassificationId() != null
+                        && def.getClassification().getKey() != null)
+            .collect(
+                Collectors.toMap(
+                    def -> def.getClassification().getClassificationId(),
+                    def -> def.getClassification().getKey(),
+                    (existing, replacement) -> existing));
```
bruderj15: Only processes `def::getClassification`.
Therefore consider mapping to it beforehand to reduce complexity.

rest/kadai-rest-spring/src/main/java/io/kadai/classification/rest/ClassificationDefinitionController.java
```diff
@@ -136,50 +136,76 @@ private Map<String, String> getSystemIds() {
...
-      }
+    return definitionList.stream()
+        .map(def -> Map.entry(def, def.getClassification()))
+        .filter(
+            defClassEntry ->
+                hasResolvableParent(defClassEntry.getValue(), keysWithDomain, systemIds))
+        .collect(
+            Collectors.toMap(
+                defClassEntry -> assembler.toEntityModel(defClassEntry.getKey()),
+                defClassEntry -> defClassEntry.getValue().getParentKey()));
```
bruderj15: Instead of `forEach` as final operation on the previous stream, what about using a `map` with side-effect (current consumer in `forEach`)?
Then you can combine it with the next iteration.

leantios: The first stream only needs the `ClassificationRepresentationModel`, whereas we need the `ClassificationDefinitionRepresentationModel` for the returning statement

bruderj15 (approved):
This is far more readable - at the cost of an extra iteration.
As `Collection<ClassificationDefinitionRepresentationModel> definitionList` is always small, this is a great deal.

PR #892, opened 2025-04-23
Closes #889 - POC for caching tasks

### LINKED ISSUE #775: Caching total table-row-counts for pagination
Pagination with direct access to valid pages requires the pagination engine to know how many rows in the table to paginate there are.
Therefore we count the amount of rows using plain old sql:
```sql
SELECT COUNT(*) FROM FOO
```

This is quite expensive.
And we run it on every query, see [QueryPagingParameter](https://github.com/kadai-io/kadai/blob/3aa69a4a972bb8562feeac12cf5db15572d1d486/rest/kadai-rest-spring/src/main/java/io/kadai/common/rest/QueryPagingParameter.java#L72):
```java
  @Override
  public List<T> apply(Q query) {
    initPageMetaData(query);
    List<T> resultList;
    if (pageMetadata != null) {
      resultList =
          query.listPage(
              Math.toIntExact(pageMetadata.getNumber()), Math.toIntExact(pageMetadata.getSize()));
    } else {
      resultList = query.list();
    }
    return resultList;
  }

  private void initPageMetaData(Q query) {
    if (page != null) {
      long totalElements = query.count();
      long maxPages = (long) Math.ceil(totalElements / pageSize.doubleValue());
      pageMetadata = new PageMetadata(pageSize, totalElements, maxPages, page);
    }
  }
```

Caching this will yield significant performance gains.

### LINKED ISSUE #889: Second-level caching with MyBatis
MyBatis has first-level [caching](https://mybatis.org/mybatis-3/sqlmap-xml.html#cache) enabled by default - per (local) session.
We barely make anything out of this as almost all queries get their own session:
```java
UserImpl user = internalKadaiEngine.executeInDatabaseConnection(() -> userMapper.findById(finalUserId));
```
Therefore we obviously execute every query, every time.

MyBatis also provides an opt-in second-level cache (cross-session), usually per mapper.
Customizable in terms of strat, ttl, and the other suspects.

Configuration is trivial:
```java
import org.apache.ibatis.cache.CacheNamespace;

@CacheNamespace(implementation = PerpetucalCache.class, eviction = LRUCache.class, size = 1024, ...)
public interface UserMapper {
  // ...
}
```

Little work, significant gains.
Although probably a no-brainer: validate with real-world usage-patterns.

bruderj15:
## Definition of Done

  - will be squashed on merge
- [ ] If extra release-notes are required, they're listed indented below:

### DIFF
--- a/lib/kadai-core/src/main/java/io/kadai/task/internal/TaskQueryImpl.java
+++ b/lib/kadai-core/src/main/java/io/kadai/task/internal/TaskQueryImpl.java
@@ -2021,17 +2021,22 @@ public class TaskQueryImpl implements TaskQuery {
   }
 
   public List<TaskSummary> list() {
-    return kadaiEngine.executeInDatabaseConnection(
-        () -> {
-          checkForIllegalParamCombinations();
-          checkOpenReadAndReadTasksPermissionForSpecifiedWorkbaskets();
-          setupJoinAndOrderParameters();
-          setupAccessIds();
-          List<TaskSummaryImpl> tasks =
-              kadaiEngine.getSqlSession().selectList(getLinkToMapperScript(), this);
+    try {
+      kadaiEngine.openConnection();
+      checkForIllegalParamCombinations();
+      checkOpenReadAndReadTasksPermissionForSpecifiedWorkbaskets();
+      setupJoinAndOrderParameters();
+      setupAccessIds();
+      List<TaskSummaryImpl> tasks =
+          kadaiEngine.getSqlSession().selectList(getLinkToMapperScript(), this);
 
-          return taskService.augmentTaskSummariesByContainedSummariesWithPartitioning(tasks);
-        });
+      // Stores read data in (transactional) cache to avoid modification (below) before caching
+      kadaiEngine.getSqlSession().commit();
+
+      return taskService.augmentTaskSummariesByContainedSummariesWithPartitioning(tasks);
+    } finally {
+      kadaiEngine.returnConnection();
+    }
   }
 
   @Override
--- a/common/kadai-common/src/main/java/io/kadai/common/internal/util/Pair.java
+++ b/common/kadai-common/src/main/java/io/kadai/common/internal/util/Pair.java
@@ -18,9 +18,10 @@
 
 package io.kadai.common.internal.util;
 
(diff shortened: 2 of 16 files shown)


bruderj15:
The test fails because [`taskService.augmentTaskSummariesByContainedSummariesWithPartitioning(tasks)`](https://github.com/kadai-io/kadai/blob/308f2d672e3bc0a8c0a12b110a99c17985df04c1/lib/kadai-core/src/main/java/io/kadai/task/internal/TaskQueryImpl.java#L2033) is not idempotent.
This shouldn't be a problem.
In fact it wasn't if the cached data was immutable as stated in the MyBatis-docs, see https://github.com/mybatis/mybatis-3/issues/3459.

lib/kadai-core/src/main/java/io/kadai/task/internal/TaskQueryImpl.java
```diff
@@ -2021,17 +2021,22 @@ public TaskQuery orderByOwnerLongName(SortDirection sortDirection) {
...
+      kadaiEngine.openConnection();
+      checkForIllegalParamCombinations();
+      checkOpenReadAndReadTasksPermissionForSpecifiedWorkbaskets();
+      setupJoinAndOrderParameters();
+      setupAccessIds();
+      List<TaskSummaryImpl> tasks =
+          kadaiEngine.getSqlSession().selectList(getLinkToMapperScript(), this);
+
+      // Stores read data in (transactional) cache to avoid modification (below) before caching
+      kadaiEngine.getSqlSession().commit();
```
bruderj15, line 2034: I don't like this at all!
Screams to cache on service-layer (not data).

bruderj15:
Tests fail because cache invalidation isn't correct.
We do not always invalidate when necessary - due to the scattered nature of database writes, e.g.:
[`TaskUpdatePriorityBatchStatement`](https://github.com/kadai-io/kadai/blob/011aeebb8aa574c27014f686d53da4fb25308472/lib/kadai-core/src/main/java/io/kadai/task/internal/jobs/helper/TaskUpdatePriorityBatchStatement.java#L28)

See issue #915.

bruderj15:
won't be realized that way due to multiple JVMs targetting the same database

PR #926, opened 2025-04-28
Closes #890 - Refine error handling in `TaskServiceImpl::createTask`

### LINKED ISSUE #890: Refine error handling in `TaskServiceImpl::createTask`
In [`TaskServiceImpl::createTask`](https://github.com/kadai-io/kadai/blob/c62009a76d601349db34df94848dd5c3629d873e/lib/kadai-core/src/main/java/io/kadai/task/internal/TaskServiceImpl.java#L555), the error handling should be refined. The `TaskAlreadyExistException` is thrown as soon as the error message contains "violation" or similar (e.g. "verletzt"). The check for `external_id` is superfluous since the SQL statement is part of the error message.

This leads to a situation where this 
```
 Cause: org.postgresql.util.PSQLException: FEHLER: NULL-Wert in Spalte »is_reopened« von Relation »task« verletzt Not-Null-Constraint
....
### The error occurred while setting parameters
### SQL: INSERT INTO TASK(ID, EXTERNAL_ID, ....
```
error resulted in a `TaskAlreadyExistException`, which is highly misleading.

leantios:
## Definition of Done

  - will be squashed on merge

### DIFF
--- a/lib/kadai-core/src/main/java/io/kadai/task/internal/TaskServiceImpl.java
+++ b/lib/kadai-core/src/main/java/io/kadai/task/internal/TaskServiceImpl.java
@@ -472,7 +472,7 @@ public class TaskServiceImpl implements TaskService {
       }
 
       Classification classification =
-              getClassificationByKeyAndDomain(task.getClassificationKey(), workbasket.getDomain());
+          getClassificationByKeyAndDomain(task.getClassificationKey(), workbasket.getDomain());
       task.setClassificationSummary(classification.asSummary());
 
       ObjectReferenceImpl.validate(task.getPrimaryObjRef(), "primary ObjectReference", "Task");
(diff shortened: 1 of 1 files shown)


bruderj15 (commented):
Can you add some tests?

Ideally integration: the constraint gets actually violated when attempting to insert.

leantios:
> Can you add some tests?
> 
> Ideally integration: the constraint gets actually violated when attempting to insert.

There exists a test which violates the external id constraint: `CreateTaskAccTest::should_ThrowException_When_CreatingTheSameTaskTwice`

Writing a test that violates another constraint (which should not raise the `TaskAlreadyExistException`) is quite hard, as in the `TaskServiceImpl` multiple checks are performed.

In my use case, I violated the Non-Null constraint of a column because I used a new database scheme with an old Kadai version. In order to reproduce, you would need to create a new column with a non-null constraint in order to have something to violate.

bruderj15:
Then consider unit-tests where you mock the exception and its message (for each database).

bruderj15 (approved):
> Then consider unit-tests where you mock the exception and its message (for each database).

After a short talk we agreed that the mentioned and already existing acceptance test for creating a duplicate task is enough.

PR #949, opened 2025-05-01
Closes #992 - Add index on workbasket-id in task-table

### LINKED ISSUE #992: Add index on workbasket-id in task-table

itscaesar:
## Definition of Done

  - will be squashed on merge


bruderj15 (commented):
Thanks for the contribution!

This looks all fine.
Have you actually tested running the migration script locally?

itscaesar:
> This looks all fine. Have you actually tested running the migration script locally?

I tested the commands for the postgres migration. I did not test for DB2. For the DB2 command I copied existing "create indexes"-commands and adapting accordingly.

bruderj15 (changes requested):
I was wondering why the CI wasn't running in here.
I opened #993 to verify that the CI succeeds there. 
This randomly also triggered the CI here ?! (We just found a GitHub bug)

More important: We also found that this approach ain't cutting it yet, see [CI failure](https://github.com/kadai-io/kadai/actions/runs/14935570058/job/41962161983#step:7:933).
Can you figure it out or may I help you? :)

itscaesar:
> More important: We also found that this approach ain't cutting it yet, see [CI failure](https://github.com/kadai-io/kadai/actions/runs/14935570058/job/41962161983#step:7:933). Can you figure it out or may I help you? :)

I checked it out and the error comes from the "create index _IF NOT EXISTS_" expression in the DB2 script (a postgres specific thing). So I deleted this, so it should work now

bruderj15 (approved):
> I checked it out and the error comes from the "create index _IF NOT EXISTS_" expression in the DB2 script (a postgres specific thing). So I deleted this, so it should work now

It does! Great job.
Whether by intuition or randomly, you were on the right track with the `...IF NOT EXISTS...` part as migration-scripts should always be idempotent.

Therefore I opened #997.
If you're interested, let me know and go for it! :)

PR #676, opened 2025-03-10
Closes #545 - Refactoring Event-Management

bruderj15:
## Definition of Done

- [x] The corresponding ticket is in state `In Review`
- [x] The corresponding ticket is attached to this Pull-Request
- [x] The commit-message-format `Closes #ISSUE_ID - PROBLEM/SOLUTION` 
  - is present as single-commit already or
  - will be squashed on merge
- [x] The changes pass the SonarQubeCloud-Quality-Gate
- [x] If the documentation needs an update, following there's a link to it:

PR #994, opened 2025-05-09
Closes #545 - Refactoring Event-Management

### LINKED ISSUE #545: Refactoring `HistoryEventManager` and `KadaiHistory`
A few observations on event-management in Kadai:
- [HistoryEventManager](https://github.com/kadai-io/kadai/blob/master/lib/kadai-core/src/main/java/io/kadai/spi/history/internal/HistoryEventManager.java) overloads `createEvent(Arg arg)` for three different base-types of Events (`TaskHistoryEvent`, ...) - their implementation is identical
- [KadaiHistory](https://github.com/kadai-io/kadai/blob/cc371dd4efdabd69c4aabc07379603d57122b69f/lib/kadai-core/src/main/java/io/kadai/spi/history/api/KadaiHistory.java) repeats that in the specification of `create(Arg arg)`
- Standard-Implementation [SimpleHistoryServiceImpl](https://github.com/kadai-io/kadai/blob/cc371dd4efdabd69c4aabc07379603d57122b69f/history/kadai-simplehistory-provider/src/main/java/io/kadai/simplehistory/impl/SimpleHistoryServiceImpl.java) implements these almost identical as well - only difference is the choice of the MyBatis-Mapper for inserting the event into the database
- Logging-Implementation [LogfileHistoryServiceImpl](https://github.com/kadai-io/kadai/blob/master/history/kadai-loghistory-provider/src/main/java/io/kadai/loghistory/impl/LogfileHistoryServiceImpl.java) does the exact same for all variants of `create(Arg arg)`
- Logging-Implementation [LogfileHistoryServiceImpl](https://github.com/kadai-io/kadai/blob/master/history/kadai-loghistory-provider/src/main/java/io/kadai/loghistory/impl/LogfileHistoryServiceImpl.java) throws `UnsupportedOperationException` for another method specified in `KadaiHistory`: `void deleteHistoryEventsByTaskIds(List<String> taskIds);`
- Context-classes that create events do so through `HistoryEventManager::createEvent`

What went wrong (I guess):
- Overloaded `create(Arg arg)` was introduced to allow static selection of MyBatis-Mapper
- Then it got required to delete tasks, so `deleteHistoryEventsByTaskIds(..)` got added to `KadaiHistory`

What this means:
- All context-classes can create events of any kind, not just their specific ones
- All context-classes can delete Task-Events
- Copy-and-paste implementations and tests for creating events
- Misused [LogfileHistoryServiceImpl](https://github.com/kadai-io/kadai/blob/master/history/kadai-loghistory-provider/src/main/java/io/kadai/loghistory/impl/LogfileHistoryServiceImpl.java) does not fulfill the specification of the interface it implements

EDIT: Find the best solution in my latest comment
What I propose to address the issues:
- Introduce interface `KadaiEvent`:
```java
public interface KadaiEvent {
  Instant getCreated();
  void setCreated(Instant createdInstant);
}
```
- Get rid of `initialize(..)` in `KadaiHistory` as this belongs to the implementation
- Make `KadaiHistory` polymorph in the event it handles and add more general logic like `delete` and co. 
I personally prefer an interface per C/R/U/D (this shouts for more abstraction over all services) - also solves `LogfileHistoryServiceImpl`-Problem:
```java
public interface KadaiHistory<T> 
  void create(T event);
  int delete(T event);
  ...
}
```
- Provide explicit implementations for all base-types of events  (`TaskHistoryEvent`, ...)
- In context-classess replace access via `HistoryEventManager` with access to specific KadaiHistory, e.g. `KadaiHistory<TaskHistoryEvent>` in `TaskServiceImpl`
- Provide access to all explicit handlers in `HistoryEventManager` via getter

What this does:
- Higher cohesion for specific Kadai-History-Implementations
- Narrower access in context-classes
- Therefore less complexity and easier extendability

But: Keeps highly similar code
- Adressing this would work with less specific `create(KadaiEvent event)` and dynamic MyBatis-Mapper selection
- However this loses out on flexibility

bruderj15:
## Definition of Done

  - will be squashed on merge
- [x] If the documentation needs an update, following there's a link to it: added entirely new docs - https://github.com/kadai-io/kadai-doc/pull/180
- [ ] If extra release-notes are required, they're listed indented below: (adding later)


history/kadai-simplehistory-provider/src/main/java/io/kadai/simplehistory/classification/internal/ClassificationHistoryServiceImpl.java
```diff
@@ -0,0 +1,124 @@
...
+    if (event.getCreated() == null) {
+      event.setCreatedNow();
+    }
+    try {
+      kadaiEngine.openConnection();
+      eventMapper.insert(event);
+    } finally {
+      kadaiEngine.returnConnection();
+    }
+    return event;
```
grittaweisheit: Maybe move this inside the try block to make it more explicit that it is only returned on success.

history/kadai-simplehistory-provider/src/main/java/io/kadai/simplehistory/workbasket/internal/WorkbasketHistoryServiceImpl.java
```diff
@@ -0,0 +1,122 @@
...
+      throw new SystemException(e.getMessage());
+    }
+    try {
+      SqlSession sqlSession = (SqlSession) sessionManager.get(kadaiEngine);
+      if (!sqlSession
+          .getConfiguration()
+          .getMapperRegistry()
+          .hasMapper(WorkbasketHistoryEventMapper.class)) {
+        sqlSession.getConfiguration().addMapper(WorkbasketHistoryEventMapper.class);
+      }
```
grittaweisheit, line 106: This is duplicated a lot considering the other service implementations. Maybe it would make sense to extract this as a helper function?

history/kadai-simplehistory-provider/src/main/java/io/kadai/simplehistory/workbasket/internal/WorkbasketHistoryServiceImpl.java
```diff
@@ -0,0 +1,122 @@
...
+          KadaiEngineImpl.class.getDeclaredField("internalKadaiEngineImpl");
+      internalKadaiEngineImpl.setAccessible(true);
+      this.kadaiEngine = (InternalKadaiEngine) internalKadaiEngineImpl.get(kadaiEngine);
+      sessionManager = KadaiEngineImpl.class.getDeclaredField("sessionManager");
+      sessionManager.setAccessible(true);
+    } catch (NoSuchFieldException e) {
+      throw new SystemException("SQL Session could not be retrieved. Aborting Startup");
+    } catch (IllegalAccessException e) {
+      throw new SystemException(e.getMessage());
+    }
```
grittaweisheit, line 98: This is common in all service implementations. Does it make sense to maybe generalize or extract it to reduce code duplication?

grittaweisheit (commented):
Looks a lot cleaner and clearer now! 
I just left a few suggestions to maybe reduce code duplication.

history/kadai-simplehistory-provider/src/main/java/io/kadai/simplehistory/workbasket/internal/WorkbasketHistoryServiceImpl.java
bruderj15: Unfortunately requires usages of Reflection.
Abstracting into e.g. a shared super abstract class is awkward.
One could provide the mappers as type-args but would later need to reify them to gain the class object.
Java's type-system is just not expressive enough.
We should keep the redundancy.

history/kadai-simplehistory-provider/src/main/java/io/kadai/simplehistory/workbasket/internal/WorkbasketHistoryServiceImpl.java
bruderj15: same argument as for general abstraction comment below

history/kadai-simplehistory-provider/src/main/java/io/kadai/simplehistory/workbasket/internal/WorkbasketHistoryServiceImpl.java
grittaweisheit: Good point!

grittaweisheit (dismissed):

common/kadai-common/src/main/java/io/kadai/common/internal/util/CollectionUtil.java
```diff
@@ -44,4 +45,16 @@ public static <T> Collection<List<T>> partitionBasedOnSize(
...
+  /**
+   * Appends a list to another.
+   *
+   * @param a the list to be appended to
+   * @param b the list to append
+   * @param <T> the type of elements in this list
+   * @return the appended list
+   */
+  public static <T> List<T> append(List<T> a, List<T> b) {
+    return Stream.concat(a.stream(), b.stream()).toList();
```
gitgoodjhe: This returns an immutable list. In your usecase this is fine but as a generic "util" this might cause problems when trying to add items after the append. What do you think?

history/kadai-simplehistory-provider/src/main/java/io/kadai/simplehistory/classification/api/ClassificationHistoryService.java
```diff
@@ -0,0 +1,64 @@
...
+import io.kadai.spi.history.api.events.classification.ClassificationHistoryEvent;
+import io.kadai.spi.history.api.exceptions.ClassificationHistoryEventNotFoundException;
+
+/**
+ * The ClassificationHistoryService manages all operations on {@linkplain ClassificationHistoryEvent
+ * ClassificationHistoryEvents}.
+ */
+public interface ClassificationHistoryService {
+
+  // region CREATE
```
gitgoodjhe: do we really need these regions considering there is only 1 method for each region currently (and probably will never be more)? Same for all the other classes with only 1 method per "region"

history/kadai-simplehistory-provider/src/main/java/io/kadai/simplehistory/task/internal/TaskDeletedHistoryEventConsumerOnCascadeDeleteEvents.java
```diff
@@ -0,0 +1,77 @@
...
+  private TaskHistoryServiceImpl taskHistoryService;
+
+  @Override
+  public void consume(TaskDeletedEvent event) {
+    if (kadaiEngine.getConfiguration().isDeleteHistoryEventsOnTaskDeletionEnabled()) {
+      final String taskEventId = event.getTaskId();
+      try {
+        taskHistoryService.deleteTaskHistoryEventsByTaskId(taskEventId);
+      } catch (NotAuthorizedException e) {
+        throw new SystemException(
```
gitgoodjhe: optional boyscouting: remove String concatenation with formatting in all cases

history/kadai-simplehistory-provider/src/test/java/io/kadai/simplehistory/classification/ClassificationHistoryServiceImplTest.java
```diff
@@ -0,0 +1,87 @@
...
+import java.util.List;
+import org.apache.ibatis.session.SqlSession;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.InjectMocks;
+import org.mockito.Mock;
+import org.mockito.Spy;
+import org.mockito.junit.jupiter.MockitoExtension;
+
+/** Unit Test for SimpleHistoryServiceImplTest. */
```
gitgoodjhe: ClassificationHistoryServiceImplTest != SimpleHistoryServiceImplTest

history/kadai-simplehistory-provider/src/test/java/io/kadai/simplehistory/task/TaskHistoryServiceImplTest.java
```diff
@@ -48,25 +43,13 @@
 
 /** Unit Test for SimpleHistoryServiceImplTest. */
```
gitgoodjhe: TaskHistoryServiceImplTest != SimpleHistoryServiceImplTest

history/kadai-simplehistory-provider/src/test/java/io/kadai/simplehistory/workbasket/WorkbasketHistoryServiceImplTest.java
```diff
@@ -0,0 +1,81 @@
...
+import java.util.List;
+import org.apache.ibatis.session.SqlSession;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.InjectMocks;
+import org.mockito.Mock;
+import org.mockito.Spy;
+import org.mockito.junit.jupiter.MockitoExtension;
+
+/** Unit Test for SimpleHistoryServiceImplTest. */
```
gitgoodjhe: WorkbasketHistoryServiceImplTest != SimpleHistoryServiceImplTest

lib/kadai-core/src/main/java/io/kadai/common/api/KadaiInitializable.java
```diff
@@ -0,0 +1,34 @@
...
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ *
+ *
+ */
+
+package io.kadai.common.api;
+
+/** Interface specifying an operation for initialization of this object during Kadai startup. */
+public interface KadaiInitializable {
```
gitgoodjhe, line 22: I like the idea but: 

-Do we really want this as a public interface? Currently it is only used in the internal classes and the SPI providers which makes sense...but this basically offers it up as a public API and could be used everywhere 

-We discussed not giving the engine to every SPI

lib/kadai-core/src/main/java/io/kadai/spi/history/internal/KadaiEventBroker.java
```diff
@@ -0,0 +1,164 @@
...
+   *
+   * @param <T> the type of the event to forward
+   * @param event the event to forward
+   */
+  @SuppressWarnings("unchecked") // Safe cast by definition of 'Object::getClass'
+  public <T extends KadaiEvent> void forward(T event) {
+    getConsumers((Class<T>) event.getClass())
+        .forEach(
+            consumer -> {
+              consumer.consume(event);
```
gitgoodjhe: are we missing a try/catch here? what happens if something goes wrong? will any runtime exception bubble out and stop forwarding to the remaining consumers?

lib/kadai-core/pom.xml
```diff
@@ -14,6 +14,7 @@
     <relativePath>../pom.xml</relativePath>
   </parent>
 
+
```
gitgoodjhe: remove please

lib/kadai-core/src/main/java/io/kadai/spi/history/internal/KadaiEventBroker.java
```diff
@@ -0,0 +1,164 @@
...
+   *
+   * @param mostSpecificClazz the class of the event of the most specific consumers to return
+   * @param <T> the type of event of the most specific consumers to return
+   * @return the list of all consumers who can consume events of the given class
+   */
+  // TODO: Although not inherently expensive, one should consider caching this
+  @SuppressWarnings({
+    "unchecked", // Safe cast by construction of 'consumers'
+  })
+  public <T extends KadaiEvent> List<KadaiEventConsumer<? super T>> getConsumers(
```
gitgoodjhe, line 122: Just a thought: with this we get a predictable type hierarchy order, but not necessarily a predictable consumer order. Should we think about this? Should we atleast log an info if there are two or more consumers for the "mostSpecificClazz"?

gitgoodjhe (changes requested):

gitgoodjhe:
I enjoyed this pull request, nice work! :) also glad you removed the database changes - was really wondering why they are in this PR (and generally)

lib/kadai-core/src/main/java/io/kadai/common/api/KadaiInitializable.java
bruderj15: > Do we really want this as a public interface?

I think yes. And you actually gave the explanation: _Currently it is [...] used in [...] the SPI providers [...]._
These are inherently public by definition. So should all parts of their spec be.

I don't think having this public is bad at all. Power is still in the hands of him invoking the method - not him implementing it.

> We discussed not giving the engine to every SPI

Yes big supporter of this as well.
But this PR doesn't change visibility at all. Only those SPIs that already declared `initialize(..)` extend `KadaiInitializable` now.

lib/kadai-core/src/main/java/io/kadai/spi/history/internal/KadaiEventBroker.java
bruderj15: > Should we atleast log an info if there are two or more consumers for the "mostSpecificClazz"?

Can't see why this would be beneficial.

> Just a thought: with this we get a predictable type hierarchy order, but not necessarily a predictable consumer order.

As-is already keeps the order, not by spec but impl 
- ArrayList of consumers per most-specific class is stable
- `ArrayList::forEach` respects order by impl
- Class-Upward-Traversion is stable as `Class::getSuperclass` is linear (no branching)
- Constructed ArrayList respects insertion-order

If we wanted to truly force a stable order at the spec we would need to return an ordered collection, e.g. `TreeSet` (basically no overhead performance-wise at this tiny amount of consumers)

bruderj15:
I united the task-consumers due to [this concern](https://github.com/kadai-io/kadai/pull/676#issuecomment-2715798561) I once raised.
Almost forgot that. 
Thanks for the order-discussion.

lib/kadai-core/src/main/java/io/kadai/common/api/KadaiInitializable.java
gitgoodjhe: okay, fair

gitgoodjhe (approved):

PR #1068, opened 2025-05-28
Closes #997 - Make migration-script for 9.3.0 -> 10.0.0 idempotent

### LINKED ISSUE #997: Make migration-script 9.3.0 -> 10.0.0 idempotent

bruderj15:
## Definition of Done

  - will be squashed on merge


leantios:
Did you try out running the migration scripts twice?

bruderj15:
> Did you try out running the migration scripts twice?

I did not run them at all, whoops.
Feel free to do that :)

leantios:
For the postgres script, apparently `version` does not have the unique constraint so `on conflict` cannot be used. I get the error:
```
SQL Error [42P10]: FEHLER: es gibt keinen Unique-Constraint oder Exclusion-Constraint, der auf die ON-CONFLICT-Angabe passt
```

common/kadai-common/src/main/resources/sql/postgres/kadai_9.3.0_to_kadai_10.0.0_schema_update_postgres.sql
```diff
@@ -1,19 +1,50 @@
...
 
-ALTER TABLE WORKBASKET_HISTORY_EVENT RENAME COLUMN ORGLEVEL_1 TO ORG_LEVEL_1;
-ALTER TABLE WORKBASKET_HISTORY_EVENT RENAME COLUMN ORGLEVEL_2 TO ORG_LEVEL_2;
-ALTER TABLE WORKBASKET_HISTORY_EVENT RENAME COLUMN ORGLEVEL_3 TO ORG_LEVEL_3;
-ALTER TABLE WORKBASKET_HISTORY_EVENT RENAME COLUMN ORGLEVEL_4 TO ORG_LEVEL_4;
+DO $$
+    BEGIN
+        ALTER TABLE WORKBASKET_HISTORY_EVENT RENAME COLUMN ORGLEVEL_1 TO ORG_LEVEL_1;
+    EXCEPTION
+        WHEN undefined_column THEN RAISE NOTICE 'Column ORGLEVEL_1 does not exist';
```
leantios, line 21: Would it not make more sense to comment 'Column ORGLEVEL_1 has already been renamed'?

bruderj15:
> For the postgres script, apparently `version` does not have the unique constraint so `on conflict` cannot be used. I get the error:
> 
> ```
> SQL Error [42P10]: FEHLER: es gibt keinen Unique-Constraint oder Exclusion-Constraint, der auf die ON-CONFLICT-Angabe passt
> ```

We should add the Unique-Constraint, shouldn't we?

leantios:
> We should add the Unique-Constraint, shouldn't we?

Yes, I agree. Multiple times the same version with different timestamps does not make any sense. I don't know if we even need the ID, theoretically, the version itself could be the key.

common/kadai-common/src/main/resources/sql/db2/kadai_9.3.0_to_kadai_10.0.0_schema_update_db2.sql
```diff
@@ -1,19 +1,100 @@
...
-ALTER TABLE TASK ADD COLUMN IS_REOPENED SMALLINT NOT NULL DEFAULT 0;
+BEGIN
+    DECLARE column_exists INT;
+    SELECT COUNT(*) INTO column_exists
+    FROM SYSCAT.COLUMNS
+    WHERE TABSCHEMA = CURRENT SCHEMA
+      AND TABNAME = 'TASK'
+      AND COLNAME = 'IS_REOPENED';
+    IF column_exists = 0 THEN
+        EXECUTE IMMEDIATE 'ALTER TABLE TASK ADD COLUMN IS_REOPENED SMALLINT NOT NULL DEFAULT 0';
```
leantios, line 13: Here, I get the error
```
Error occurred during SQL script execution

Reason:
SQL Error [42601]: An unexpected token "END-OF-STATEMENT" was found following "AULT 0'".  Expected tokens may include:  " 
    END IF".. SQLCODE=-104, SQLSTATE=42601, DRIVER=4.33.3
```
This is a syntax error, but I could not find out where it comes from

leantios (changes requested):

bruderj15:
won't fix

PR #1122, opened 2025-06-06
Closes #1121 - Fix exception handling in task distribution

### LINKED ISSUE #1121: Fix Bug in Task Distribution
When distributing multiple tasks, it can happen that some tasks cause issues. In the transferTasks method, this meant that the remaining tasks were successfully transferred, and a result was returned that included the problematic tasks along with the corresponding exceptions. In contrast, the new distributeTasks method aborts the entire operation. Specifically, this concerns the TaskNotFoundException. The new behavior could lead to significant delays in operations and/or increased effort.

Bug refers to: https://github.com/kadai-io/kadai/issues/372

jannikFuellgrafEnvite:
## Definition of Done

  - will be squashed on merge

### DIFF
--- a/lib/kadai-core/src/main/java/io/kadai/task/internal/TaskDistributor.java
+++ b/lib/kadai-core/src/main/java/io/kadai/task/internal/TaskDistributor.java
@@ -22,6 +22,7 @@ import io.kadai.common.api.BulkOperationResults;
 import io.kadai.common.api.exceptions.InvalidArgumentException;
 import io.kadai.common.api.exceptions.KadaiException;
 import io.kadai.common.internal.InternalKadaiEngine;
+import io.kadai.common.internal.util.Pair;
 import io.kadai.spi.task.api.TaskDistributionProvider;
 import io.kadai.spi.task.internal.TaskDistributionManager;
 import io.kadai.task.api.exceptions.TaskNotFoundException;
@@ -33,6 +34,7 @@ import io.kadai.workbasket.api.exceptions.NotAuthorizedOnWorkbasketException;
 import io.kadai.workbasket.api.exceptions.WorkbasketNotFoundException;
 import io.kadai.workbasket.api.models.Workbasket;
 import io.kadai.workbasket.api.models.WorkbasketSummary;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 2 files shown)


lib/kadai-core/src/main/java/io/kadai/task/internal/TaskDistributor.java
```diff
@@ -170,7 +154,7 @@ private BulkOperationResults<String, KadaiException> distributeTasks(
 
       workbasketService.checkAuthorization(sourceWorkbasketId, WorkbasketPermission.DISTRIBUTE);
 
-      taskIds = resolveTaskIds(sourceWorkbasketId, taskIds);
+      taskIds = resolveTaskIds(sourceWorkbasketId, taskIds, operationResults);
```
bruderj15: Instead of this C++ Multi-Return-Value approach I'd personally favor a more functional one, returning a `Pair<List<String>, BulkOperationResults<String, KadaiException>>`.
This significantly reduces cognitive load.

Similar to `transferTasks(newTaskDistribution, taskTransferFlags, operationResults);` a few lines down.
Instead of returning `void` it should really return the `operationResults` - you then merge them afterwards.

lib/kadai-core/src/main/java/io/kadai/task/internal/TaskDistributor.java
```diff
@@ -160,8 +145,7 @@ private BulkOperationResults<String, KadaiException> distributeTasks(
       Map<String, Object> additionalInformation)
       throws InvalidArgumentException,
           NotAuthorizedOnWorkbasketException,
-          WorkbasketNotFoundException,
-          TaskNotFoundException {
+          WorkbasketNotFoundException {
 
     BulkOperationResults<String, KadaiException> operationResults = new BulkOperationResults<>();
```
bruderj15: Reader might wonder why this needs to live outside the try-catch-block.

bruderj15 (commented):
Help me with the desired behavior:
- Partial failure for non-existing tasks
- Entire failure for non-existing workbaskets?

If the latter is actually desired, revisit `transferTasks()`.
It currently wraps the `WorkbasketNotFoundException` inside the `BulkOperationResults`.

bruderj15 (approved):

PR #1204, opened 2025-07-04
Closes #1194 - Distinguish error-handling for bulk-imports

### LINKED ISSUE #1194: Distinguish error-handling for bulk-imports
Follow-Up to #855 

## Description

Currently some API-Elements - primarily REST-Endpoints for Bulk-Imports - require all items in the bulk input to be unique.
If there are any _logical_ duplicates (some identical key), it fails with a `409 CONFLICT` and error-code e.g. `CLASSIFICATION_ALREADY_EXISTS`.

Http- and Error-Code are both misleading.
The semantics of the latter are overloaded with use cases where insertion of an item into the database fails because the item actually already exists.

Therefore, add a new Error-Type and -Code.

Currently, below occurences are known:
- `ClassificationDefinitionController::importClassifications`
- `WorkbasketDefinitionController::importWorkbaskets`
- `WorkbasketService::setWorkbasketAccessItems`

## Definition of Done

- [x] Checked for other occurences of the misleading usage of `[Entity]AlreadyExistException`
- [x] Added a new exception for invalid payloads due to logical duplicates
  - [x] Added Java-Type
  - [x] Added Error-Code
  - [x] Added Contract-Test for Error-Code
- [x] Affected REST-Endpoints now respond with `400 BAD_REQUEST` if the payload contains _logical_ duplicates
- [x] Affected REST-Endpoints now respond with the new Error-Code if the payload contains _logical_ duplicates
- [x] Added tests for the new behaviour & adjusted affected tests

bruderj15:
## Definition of Done

  - will be squashed on merge
  - Fixed bug where duplicates in REST-Api-Payloads would return a `409 CONFLICT` with error-code `[ENTITY]_ALREADY_EXIST_EXCEPTION`
  - The affected endpoints now return a `400 BAD_REQUEST` with error-code `LOGICAL_DUPLICATE_IN_PAYLOAD`
  - This affects 
    - `POST /api/v1/workbasket-definitions`
    - `POST /api/v1/classification-definitions`
    - `PUT /api/v1/workbaskets/{workbasketId}/workbasketAccessItems`


bruderj15:
SonarQube stoopid, `KadaiRestExceptionHandler` is covered via Int-Tests

common/kadai-common/src/main/java/io/kadai/common/api/exceptions/LogicalDuplicateInPayloadException.java
```diff
@@ -0,0 +1,43 @@
...
+import java.util.Map;
+
+/**
+ * This exception is thrown when the payload for a request to a REST-Endpoint contains logical
+ * duplicates.
+ */
+public class LogicalDuplicateInPayloadException extends KadaiException {
+
+  public static final String ERROR_KEY = "LOGICAL_DUPLICATE_IN_PAYLOAD";
+  private final Serializable duplicate;
```
leantios, line 31: Why do you use Serializable here and not String? As far as I am concerned, all the usages are with String

rest/kadai-rest-spring/src/main/java/io/kadai/workbasket/rest/WorkbasketDefinitionApi.java
```diff
@@ -153,15 +157,12 @@ ResponseEntity<WorkbasketDefinitionCollectionRepresentationModel> exportWorkbask
...
-            description =
-                "WORKBASKET_ALREADY_EXISTS, WORKBASKET_ACCESS_ITEM_ALREADY_EXISTS, "
-                    + "ENTITY_NOT_UP_TO_DATE",
+            description = "WORKBASKET_ACCESS_ITEM_ALREADY_EXISTS, ENTITY_NOT_UP_TO_DATE",
             content = {
               @Content(
                   schema =
                       @Schema(
                           anyOf = {
-                            WorkbasketAlreadyExistException.class,
```
leantios, line 164: Now what happens if the key is the same but the domain not? Then a new Workbasket is created with the same key but different domain? Should we not throw a WorkbasketAlreadyExistException?

rest/kadai-rest-spring/src/main/java/io/kadai/workbasket/rest/WorkbasketDefinitionController.java
```diff
@@ -152,7 +151,12 @@ public ResponseEntity<Void> importWorkbaskets(@RequestParam("file") MultipartFil
 
         newId = systemIds.get(logicalId(importedWb));
       } else {
-        newId = workbasketService.createWorkbasket(wbWithoutId).getId();
+        try {
+          newId = workbasketService.createWorkbasket(wbWithoutId).getId();
+        } catch (WorkbasketAlreadyExistException ignore) {
+          // Cannot exist because we previously checked and then updated
```
leantios, line 157: See above: didn't we only check for equality of both key and domain? So the same key could still be passed but with different domain

rest/kadai-rest-spring/src/main/java/io/kadai/classification/rest/ClassificationDefinitionApi.java
```diff
@@ -114,14 +117,13 @@ ResponseEntity<ClassificationDefinitionCollectionRepresentationModel> exportClas
...
             responseCode = "409",
-            description = "ENTITY_NOT_UP_TO_DATE, CLASSIFICATION_ALREADY_EXISTS",
+            description = "ENTITY_NOT_UP_TO_DATE",
             content = {
               @Content(
                   schema =
                       @Schema(
                           anyOf = {
                             ConcurrencyException.class,
-                            ClassificationAlreadyExistException.class
```
leantios, line 124: Same issue as for the workbaskets: what if there is a classification with the same key but a different domain as an existing one?

leantios (changes requested):

common/kadai-common/src/main/java/io/kadai/common/api/exceptions/LogicalDuplicateInPayloadException.java
bruderj15: I prefer the most general type :)

rest/kadai-rest-spring/src/main/java/io/kadai/workbasket/rest/WorkbasketDefinitionApi.java
bruderj15: Good question. I went along with the data-model where only the combination of both is required to be unique.
I will bring it to discussion in the upcoming weekly.

https://github.com/kadai-io/kadai/blob/cf72eb0d756a54ae7f107011b19845a01e53c802/common/kadai-common/src/main/resources/sql/postgres/kadai-schema-postgres.sql#L77

bruderj15: As of the weekly: data-model is right, `key` itself does not need to be unique, only the composite with `domain`.
Therefore the implemented behavior reflects the desired.

@leantios nonetheless I'd appreciate if you'd take a thorough look on it again.

leantios (approved):
LGTM

PR #1205, opened 2025-07-07
Closes #1196 - Unify Persistence-Layer-Access to one DAO per Entity

### LINKED ISSUE #1196: Unify Persistence-Layer-Access to one DAO per Entity
Follow-Up to #915 

## Description

In general, CRUD-Applications restrict access to the persistence-layer to at most one DAO per entity.
Using MyBatis as Persistence-Framework, our DAO is the `Mapper`, e.g. `TaskMapper`.
Throughout the code-base this is done for most parts.

However, there are exceptions where some classes access the persistence-layer - bypassing the DAO.
This ships with unnecessary complexity and scatters the access to the persistence-layer.

Therefore, unify the access by moving all access to the respective DAOs.

(Query-Mappers are an exception here, do not touch them)

## Definition of Done

- [x] Checked for all occurences of non-DAO persistence-access
- [x] Moved all persistence-access to the respective DAOs

bruderj15:
Did not adjust [`TaskUpdatePriorityBatchStatement`](https://github.com/kadai-io/kadai/blob/master/lib/kadai-core/src/main/java/io/kadai/task/internal/jobs/helper/TaskUpdatePriorityBatchStatement.java#L35) because MyBatis has no configuration-driven way to execute batch-statements.

I did not want to use dynamic `<foreach>...` for the task-priority updates as this would be much more expensive than manual batching.
As caches won't be used anyways  and the importance of this ticket is degarded, we're fine with the little scattering of the data-access-layer.

## Definition of Done

  - will be squashed on merge

### DIFF
--- a/rest/kadai-rest-spring/src/main/java/io/kadai/user/jobs/UserInfoRefreshJob.java
+++ b/rest/kadai-rest-spring/src/main/java/io/kadai/user/jobs/UserInfoRefreshJob.java
@@ -29,11 +29,10 @@ import io.kadai.common.internal.transaction.KadaiTransactionProvider;
 import io.kadai.common.rest.ldap.LdapClient;
 import io.kadai.common.rest.util.ApplicationContextProvider;
 import io.kadai.spi.user.internal.RefreshUserPostprocessorManager;
-import io.kadai.task.internal.jobs.helper.SqlConnectionRunner;
 import io.kadai.user.api.exceptions.UserAlreadyExistException;
 import io.kadai.user.api.exceptions.UserNotFoundException;
 import io.kadai.user.api.models.User;
-import java.sql.PreparedStatement;
+import io.kadai.user.internal.UserServiceImpl;
 import java.time.Duration;
 import java.util.List;
 import org.slf4j.Logger;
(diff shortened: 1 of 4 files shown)


rest/kadai-rest-spring/src/main/java/io/kadai/user/jobs/UserInfoRefreshJob.java
```diff
@@ -91,24 +88,17 @@ protected void execute() {
     }
   }
 
-  private void clearExistingUsersAndGroupsAndPermissions() {
+  private void clearExistingUsersAndGroupsAndPermissions() throws NotAuthorizedException {
+    if (LOGGER.isDebugEnabled()) {
```
CRoberto1926, line 92: I think you can delete this guard and the guard on line 99. There is no String Parsing in the LOGGER, so no performance impairements.

CRoberto1926 (commented):
LGTM

rest/kadai-rest-spring/src/main/java/io/kadai/user/jobs/UserInfoRefreshJob.java
bruderj15: Afaik the log is still produced (sent to the Log-Manager) and just trashed there.
We have usages like here throughout the codebase very often.

I personally do not like it either and would pay the minimal (and certainly negligible) performance-price for readability.
To keep it consistent across the code-base, I'll leave it like this for now.

CRoberto1926 (approved):

Issue #1206, opened 2025-07-07
Consider `ExecutorType::BATCH` over `ExecutorType::SIMPLE` with dynamic `<foreach>...`


bruderj15:
All of our non-trivial (trivial = `UPDATE xy = foo WHERE state = bar`) bulk-operations use dynamic sql with MyBatis' DSLs `<foreach>...</foreach>`.
As the statement cannot be known at compilation time, it cannot be _prepared_ and therefore is dynamic and inefficient.
It's also worth noting, that this limits the size of the bulk to the maximum amount of SQL-Parameters.
See `TaskMapper` for example:

https://github.com/kadai-io/kadai/blob/fb6c035001cb496153441911e8f4012ebba09693/lib/kadai-core/src/main/java/io/kadai/task/internal/TaskMapper.java#L172-L179

Ideally this would be done with batching.
Although MyBatis does not allow configuration-driven batching (e.g. via annotation), we can enable batching at the session-level by obtaining sessions with `ExecutorType::BATCH` instead of the currently used `ExecutorType::SIMPLE`.
This results in the use of a _prepared_ statement with batching.

There is an example from MyBatis [here](https://github.com/mybatis/mybatis-3/wiki/FAQ#how-do-i-code-a-batch-insert).
Surely we would greatly benefit from batching performance-wise.

But how easy is this to integrate?

PR #1216, opened 2025-07-09
Closes #1201 - Bulk Comment Creation Endpoint

### LINKED ISSUE #1201: Bulk Comment Creation Endpoint
## Description

Introduce a new REST API endpoint to allow adding a single comment to multiple tasks in one request.

## Endpoint

- **URL:** `POST /api/v1/tasks/comments`
- **Request Body:**
```
  {
    "taskIds": ["task-id-1", "task-id-2", "..."],
    "textField": "This is a comment to be added to all tasks."
  }
```

- **Response:**
```
{
  "failedTaskIds": ["task-id-2", "..."]
}
```
## Suggestion
Use the existing Java API for creating a single task comment and loop over it.

## Acceptance Criteria (Definition of Done)
 
- [x] A new endpoint POST /api/v1/tasks/comments is implemented
- [x] Failed comment creations are collected and returned in the response
- [x]  Unit and integration tests are in place and passing
- [x]  API documentation is updated with usage examples

lucalimbach:
## Definition of Done

  - will be squashed on merge
- add new endpoint '/v1/tasks/comments'
- updated Swagger UI documentation


rest/kadai-rest-spring/src/main/java/io/kadai/task/rest/TaskCommentApi.java
```diff
@@ -406,4 +410,66 @@ ResponseEntity<TaskCommentRepresentationModel> createTaskComment(
...
+            }
+            """
+                            )
+                    )
+            )
+        }
+    )
+  @PostMapping(path = RestEndpoints.URL_MULTIPLE_TASKS_COMMENT)
+  @Transactional(rollbackFor = Exception.class)
+  ResponseEntity<Map<String, List<String>>> createTaskCommentForMultipleTasks(
```
bruderj15: We aim to sell a REST-API with (at least partial) [HATEOAS](https://de.wikipedia.org/wiki/HATEOAS)-support.

Have you considered a HATEOAS-[DTO](https://en.wikipedia.org/wiki/Data_transfer_object) instead of the `Map<..,..>`? It would include all references to the failed tasks.

rest/kadai-rest-spring/src/main/java/io/kadai/common/rest/RestEndpoints.java
```diff
@@ -86,6 +86,7 @@ public final class RestEndpoints {
       API_V1 + "tasks/distribute/{workbasketId}";
 
   // task comment endpoints
+  public static final String URL_MULTIPLE_TASKS_COMMENT = API_V1 + "tasks/comments";
```
bruderj15: Personal preference: I'd call it `URL_TASKS_COMMENT`

rest/kadai-rest-spring/src/main/java/io/kadai/task/rest/models/TaskCommentMultipleTasksRepresentationModel.java
```diff
@@ -0,0 +1,54 @@
...
+
+import io.swagger.v3.oas.annotations.media.Schema;
+import java.util.List;
+import org.springframework.hateoas.RepresentationModel;
+
+/**
+ * EntityModel class for adding a comment to multiple tasks.
+ */
+@Schema(description = "EntityModel class for adding a comment to multiple tasks.")
+public class TaskCommentMultipleTasksRepresentationModel
```
bruderj15:
```suggestion
public class TasksCommentBatchRepresentationModel
```

rest/kadai-rest-spring/src/main/java/io/kadai/task/rest/TaskCommentController.java
```diff
@@ -169,4 +173,39 @@ public ResponseEntity<TaskCommentRepresentationModel> createTaskComment(
...
+        failedTaskIds.add(taskId);
+      }
+    }
+
+    if (failedTaskIds.isEmpty()) {
+      return ResponseEntity.status(HttpStatus.CREATED).build();
+    } else {
+      Map<String, List<String>> response = new HashMap<>();
+      response.put("failedTaskIds", failedTaskIds);
+      return ResponseEntity.status(HttpStatus.MULTI_STATUS).body(response);
```
bruderj15: The `207` is the best-practice approach for Partial-Failure surely.
But it should include multiple responses, e.g.:
```json
{
  "results": [
    { "taskId": 1, "status": 200 },
    { "taskId": 2, "status": 403, "error": "NOT_AUTHORIZED_ON_TASK_COMMENT" },
    { "taskId": 3, "status": 404, "error": "TASK_NOT_FOUND" },
  ]
}
```

However, often a simple `200` even for partial success is plausible.
You should really ask the clients here (JIRA) if they want:
- `200` for partial failure or
- `207` for partial failure

rest/kadai-rest-spring/src/test/java/io/kadai/task/rest/TaskCommentControllerIntTest.java
```diff
@@ -297,6 +300,96 @@ void should_FailToCreateTaskComment_When_TaskIdIsNonExisting() {
         .isEqualTo(HttpStatus.NOT_FOUND);
   }
 
+  @Test
+  void should_CreateTaskCommentForMultipleTasks_When_TasksAreVisible() {
```
bruderj15: When is a task _visible_?

rest/kadai-rest-spring/src/main/java/io/kadai/task/rest/TaskCommentController.java
```diff
@@ -169,4 +173,39 @@ public ResponseEntity<TaskCommentRepresentationModel> createTaskComment(
...
+        TaskCommentRepresentationModel taskCommentRepresentationModel =
+                new TaskCommentRepresentationModel();
+        taskCommentRepresentationModel.setTaskId(taskId);
+        taskCommentRepresentationModel
+                .setTextField(taskCommentMultipleTasksRepresentationModel
+                        .getTextField());
+
+        TaskComment taskComment = taskCommentRepresentationModelAssembler
+                .toEntityModel(taskCommentRepresentationModel);
+        taskService.createTaskComment(taskComment);
```
bruderj15: Most of the time (I assume), creation of the comment for all tasks will succeed.
Currently we run `n` transactions for `n` tasks to create the comment for, quite expensive (think 1000 tasks in bulk).

When most of the time, all of them succeed, me may _optimistically_ try and write them all at once (one transaction).
If it succeeds, great we're done (much faster).
If not, the entire transaction failed and we will go with the 1-by-1 fallback approach as you've implemented it now.

What do you think?

bruderj15 (commented):
Looks very good already!

rest/kadai-rest-spring/src/main/java/io/kadai/task/rest/TaskCommentController.java
bruderj15: We agreed to use `200`.

rest/kadai-rest-spring/src/main/java/io/kadai/task/rest/TaskCommentApi.java
bruderj15: No hypermedia-links for now.
Probably should use [`BulkOperationResultsRepresentationModel`](https://github.com/kadai-io/kadai/blob/master/rest/kadai-rest-spring/src/main/java/io/kadai/task/rest/models/BulkOperationResultsRepresentationModel.java).

lib/kadai-core/src/main/java/io/kadai/task/api/TaskService.java
```diff
@@ -1464,6 +1464,10 @@ BulkOperationResults<String, KadaiException> deleteTasks(List<String> tasks)
   TaskComment createTaskComment(TaskComment taskComment)
       throws TaskNotFoundException, InvalidArgumentException, NotAuthorizedOnWorkbasketException;
 
+  BulkOperationResults<String, KadaiException> createTaskCommentsBulk(
```
bruderj15, line 1477: Might want to add some docs here

lib/kadai-core/src/main/java/io/kadai/task/internal/TaskCommentServiceImpl.java
```diff
@@ -140,6 +142,50 @@ TaskComment createTaskComment(TaskComment taskCommentToCreate)
...
+              return c;
+            })
+            .toList();
+
+    BulkOperationResults<String, KadaiException> errors = new BulkOperationResults<>();
+    boolean bulkSucceeded = true;
+
+    try {
+      kadaiEngine.openConnection();
+      for (TaskComment c : comments) {
```
bruderj15: As `TaskService::createTaskComment` itself starts its own transaction, we still get `n` transactions here (actually `n+1` now).

Consider adding `TaskService::createTaskComments` with an implementation that starts one transaction and then directly uses the `TaskMapper` for `n` inserts (requires some more logic - think comment-count).

This would reduce the transaction-overhead (for the optimistic path).
Obviously still not ideal as we run insert `n` times instead of one actual SQL-Batch of size `n`.
This however is related to #1206 and not in scope here.

rest/kadai-rest-spring/src/main/java/io/kadai/task/rest/TaskCommentApi.java
```diff
@@ -406,4 +409,69 @@ ResponseEntity<TaskCommentRepresentationModel> createTaskComment(
...
+   * @title Create a Task Comment for multiple Tasks
+   * @param tasksCommentBatchRepresentationModel contains the list of Task IDs and the comment text
+   * @return bulkOperationResponseModel with
+   *     HTTP200 OK and a list of Task IDs for which the comment creation failed
+   *
+   */
+  @Operation(
+        summary = "Create Task Comments for multiple Tasks",
+        description = "Creates the same Task Comment for all provided Task IDs. "
+                + "Always returns 200 OK; response body contains the task IDs with their error code"
```
bruderj15: > Always returns 200 OK

A little too dominant. Could still receive e.g. all 5xx.

rest/kadai-rest-spring/src/main/java/io/kadai/task/rest/models/BulkOperationResponseModel.java
```diff
@@ -0,0 +1,67 @@
...
+import io.kadai.common.api.exceptions.KadaiException;
+import io.swagger.v3.oas.annotations.media.Schema;
+import java.util.List;
+import java.util.Objects;
+
+/**
+ * Response DTO for a bulk operation.
+ */
+@Schema(description = "Response model for bulk operations")
+public class BulkOperationResponseModel {
```
bruderj15: I think you can just use this type: 

https://github.com/kadai-io/kadai/blob/0f9b7dc018ebc52d4c6b5799dc0f34491e01f465/rest/kadai-rest-spring/src/main/java/io/kadai/task/rest/models/BulkOperationResultsRepresentationModel.java#L27

rest/kadai-rest-spring/src/main/java/io/kadai/task/rest/models/TasksCommentBatchRepresentationModel.java
```diff
@@ -0,0 +1,54 @@
...
+
+import io.swagger.v3.oas.annotations.media.Schema;
+import java.util.List;
+import org.springframework.hateoas.RepresentationModel;
+
+/**
+ * EntityModel class for adding a comment to multiple tasks.
+ */
+@Schema(description = "EntityModel class for adding a comment to multiple tasks.")
+public class TasksCommentBatchRepresentationModel
```
bruderj15, line 27: Throughout the code-base types for similar usage look like this:

https://github.com/kadai-io/kadai/blob/0f9b7dc018ebc52d4c6b5799dc0f34491e01f465/rest/kadai-rest-spring/src/main/java/io/kadai/task/rest/models/TransferTaskRepresentationModel.java#L25

I cannot tell what's better, yours probably is valid too.
I'd suggest keeping it consistent.

bruderj15 (commented):
Looks really good!

Try to add a few more REST-Tests.
Ideally some more with partial failure, e.g. due to insufficient permissions.

rest/kadai-rest-spring/src/main/java/io/kadai/task/rest/models/BulkOperationResponseModel.java
lucalimbach: BulkOperationResultsRepresentationModel extends RepresentationModel and adds HATEOAS _links, while the earlier [requirement ](https://github.com/kadai-io/kadai/pull/1216#discussion_r2196971034)was to avoid links, so I added a simple DTO (List<id, ErrorCode>). Should the response stay as this DTO or switch to the HATEOAS model?

rest/kadai-rest-spring/src/main/java/io/kadai/task/rest/models/TasksCommentBatchRepresentationModel.java
lucalimbach: Current implementation follows [TaskCommentRepresentationModel ](https://github.com/kadai-io/kadai/blob/0f9b7dc018ebc52d4c6b5799dc0f34491e01f465/rest/kadai-rest-spring/src/main/java/io/kadai/task/rest/models/TaskCommentRepresentationModel.java)(extends RepresentationModel, uses Schema). 

For these bulk/batch models, which style should we align with? I can refactor accordingly.

rest/kadai-rest-spring/src/main/java/io/kadai/task/rest/models/BulkOperationResponseModel.java
bruderj15: Ah, my bad. The linked requirement itself is contradicting already.
Go for `BulkOperationResultsRepresentationModel` and HATEOAS.

rest/kadai-rest-spring/src/main/java/io/kadai/task/rest/models/TasksCommentBatchRepresentationModel.java
bruderj15: I'd prefer aligning this to the usage in the remainder of the code-base - following the schema `TransferTaskRepresentationModel ` provides.

lib/kadai-core/src/main/java/io/kadai/task/internal/TaskCommentServiceImpl.java
```diff
@@ -140,6 +147,54 @@ TaskComment createTaskComment(TaskComment taskCommentToCreate)
     return taskCommentImplToCreate;
   }
 
+  public BulkOperationResults<String, KadaiException> createTaskCommentsBulk(
+          List<String> taskIds, String text) throws InvalidArgumentException {
+
+    if (taskIds == null || taskIds.isEmpty()) {
```
bruderj15: I buy the Null-Check, but why is it invalid to supply an empty list?

lib/kadai-core/src/main/java/io/kadai/task/internal/TaskCommentServiceImpl.java
```diff
@@ -140,6 +147,54 @@ TaskComment createTaskComment(TaskComment taskCommentToCreate)
     return taskCommentImplToCreate;
   }
 
+  public BulkOperationResults<String, KadaiException> createTaskCommentsBulk(
+          List<String> taskIds, String text) throws InvalidArgumentException {
```
bruderj15: Type of `taskIds` can be more general, right? `Collection<String>?`

lib/kadai-core/src/main/java/io/kadai/task/internal/TaskCommentServiceImpl.java
```diff
@@ -140,6 +147,54 @@ TaskComment createTaskComment(TaskComment taskCommentToCreate)
...
+            .map(MinimalTaskSummary::getTaskId)
+            .collect(Collectors.toSet());
+
+    List<TaskCommentImpl> toInsert = new ArrayList<>();
+    for (String taskId : taskIds) {
+      try {
+        if (!existingTaskIds.contains(taskId)) {
+          taskService.getTask(taskId);
+          continue;
+        }
```
bruderj15: By spec of `TaskMapper::findExistingTasks` all existing tasks for given ids are returned.

What is 
```java
        if (!existingTaskIds.contains(taskId)) {
          taskService.getTask(taskId);
          continue;
        }
```
for then? 
Also we're ignoring the result? You're probably implying the `catch`-block with `TaskNotFoundException` here.

I think we can simplify this by
1. finding all existing tasks
2. doing the work for them
3. computing the non-existing taskIds
4. adding `TaskNotFoundException` for all of them in the Bulk-Result

lib/kadai-core/src/main/java/io/kadai/task/internal/TaskCommentServiceImpl.java
```diff
@@ -140,6 +147,54 @@ TaskComment createTaskComment(TaskComment taskCommentToCreate)
...
+    if (taskIds == null || taskIds.isEmpty()) {
+      throw new InvalidArgumentException("taskIds must not be null/empty");
+    }
+    if (text == null || text.isEmpty()) {
+      throw new InvalidArgumentException("text must not be null/empty");
+    }
+
+    BulkOperationResults<String, KadaiException> errors = new BulkOperationResults<>();
+    Instant now = Instant.now();
+
```
bruderj15: I think we should start a transaction here

lib/kadai-core/src/main/java/io/kadai/task/internal/TaskCommentServiceImpl.java
```diff
@@ -140,6 +146,51 @@ TaskComment createTaskComment(TaskComment taskCommentToCreate)
...
+      Instant now = Instant.now();
+
+      Set<String> existingTaskIds = taskMapper.findExistingTasks(taskIds, null).stream()
+              .map(MinimalTaskSummary::getTaskId)
+              .collect(Collectors.toSet());
+
+      taskIds.stream()
+              .filter(id -> !existingTaskIds.contains(id))
+              .forEach(id -> errors.addError(
+                      id, new TaskNotFoundException("Task not found: " + id)));
```
bruderj15: Careful, the constructor takes the task-id, not a human-readable message.
```suggestion
                      id, new TaskNotFoundException(id)));
```

lib/kadai-core/src/main/java/io/kadai/task/internal/TaskCommentServiceImpl.java
```diff
@@ -140,6 +146,51 @@ TaskComment createTaskComment(TaskComment taskCommentToCreate)
...
+
+      List<TaskCommentImpl> toInsert = existingTaskIds.stream()
+              .map(id -> {
+                TaskCommentImpl comment = (TaskCommentImpl) newTaskComment(id);
+                comment.setTextField(text);
+                comment.setId(null);
+                validateNoneExistingTaskCommentId(comment.getId());
```
bruderj15: This is redundant, isn't it?
```suggestion
                comment.setId(null);
```

rest/kadai-rest-spring/src/main/java/io/kadai/common/rest/RestEndpoints.java
```diff
@@ -86,6 +86,7 @@ public final class RestEndpoints {
       API_V1 + "tasks/distribute/{workbasketId}";
 
   // task comment endpoints
+  public static final String URL_TASKS_COMMENT = API_V1 + "tasks/comments";
```
bruderj15:
```suggestion
  public static final String URL_TASKS_COMMENTS = API_V1 + "tasks/comments";
```

rest/kadai-rest-spring/src/main/java/io/kadai/task/rest/TaskCommentApi.java
```diff
@@ -406,4 +409,73 @@ ResponseEntity<TaskCommentRepresentationModel> createTaskComment(
...
+   * @param tasksCommentBatchRepresentationModel list of Task IDs and the comment text
+   * @return HTTP 200 with a {@link BulkOperationResultsRepresentationModel} containing the IDs
+   *         that failed with their error codes; the list is empty if all comments were
+   *         created successfully. Otherwise, an appropriate error status (e.g. 400).
+   */
+  @Operation(
+            summary = "Create Task Comments for multiple Tasks",
+            description = "Creates the same comment for all provided task IDs. "
+                      + "Returns 200 on success with a map of failed task IDs to their error codes "
+                      + "(empty map if all succeeded)."
```
bruderj15:
```suggestion
                      + "Returns 200 on success with a map of failed task IDs to their error codes. "
```

lib/kadai-core/src/main/java/io/kadai/task/internal/TaskCommentServiceImpl.java
```diff
@@ -140,6 +146,51 @@ TaskComment createTaskComment(TaskComment taskCommentToCreate)
...
+          Collection<String> taskIds, String text) throws InvalidArgumentException {
+
+    if (taskIds == null) {
+      throw new InvalidArgumentException("taskIds must not be null");
+    }
+    if (text == null || text.isEmpty()) {
+      throw new InvalidArgumentException("text must not be null/empty");
+    }
+
+    kadaiEngine.openConnection();
```
bruderj15: I'd move this into the try-block.

NB: This is a Kadai-inherent anti-pattern. Opening a connection _should_ be done with a `try-with-resources` statement. We can only adapt here.

bruderj15 (commented):
Looks good!
Resolve the Sonar-Issues and we're ready!

rest/kadai-rest-spring/src/test/java/io/kadai/task/rest/TaskCommentControllerIntTest.java
```diff
@@ -541,4 +687,108 @@ void should_FailToDeleteTaskComment_When_TaskCommentIsNotExisting() {
         .extracting(HttpStatusCodeException::getStatusCode)
         .isEqualTo(HttpStatus.NOT_FOUND);
   }
+
+  @Test
+  void should_Equal_when_tasksWithErrorsAreSame_and_hashCodesMatch() {
```
bruderj15: This and all other tests below are Unit-Tests for the `BulkOperationResultsRepresentationModel`.
This file shall only contain Integration-Tests for the `TaskCommentController`.

Certainly valid to test the behavior with unit-tests though.
I even endorse it. However, in general Kadai doesnt really unit-test.

You could move these Unit-Tests in a dedicated file or just remove them altogether.

lib/kadai-core/src/main/java/io/kadai/task/internal/ServiceLevelHandler.java
```diff
@@ -56,18 +57,18 @@ class ServiceLevelHandler {
   private final TaskMapper taskMapper;
   private final AttachmentMapper attachmentMapper;
   private final WorkingTimeCalculator workingTimeCalculator;
-  private final TaskServiceImpl taskServiceImpl;
+  private final TaskService taskService;
```
bruderj15: Why are we doing this throughout this whole PR?
SonarQube complained?
Unfortunately it requires downstream-casting, worsening readability.

For internal classes (inside packages `...internal...`) it's okay to use the impl instead of the interface.

Or are there any reasons I might not think of, why you changed that?

lucalimbach: SonarQube flagged a cyclic dependency Issue between TaskServiceImpl and TaskCommentServiceImpl.
If you have a different solution in mind, I can roll this back.

bruderj15 (commented):
The implementation is fine.
Can you add a few lines to all the integration-tests which verify, that the comment has actually been created for the tasks?

I think you can simply inject the `Task(Comment)Service` to read the comments for a task.
Check out [`@KadaiInject`](https://github.com/kadai-io/kadai/blob/d0f32162c90663d890df900048b6878fd0a990dd/lib/kadai-test-api/src/main/java/io/kadai/testapi/KadaiInject.java#L50) and it's usages.

lucalimbach:
@bruderj15 We're using @KadaiSpringBootTest here, not @KadaiIntegrationTest, so @KadaiInject won’t work. Would it make sense to create a separate test class using @KadaiIntegrationTest specifically for this?

bruderj15 (approved):
The way you solved it with the tests certainly works.
Well done!

PR #1223, opened 2025-07-11
Closes #1222 - Fix non-existing LDAP-Entries fail saving workbaskests

### LINKED ISSUE #1222: Non-existing LDAP-Entries fail saving workbaskests
LDAP group names change frequently and LDAP users are removed. 
This leads to problems when these workbaskets need to be processed in the KADAI Admin dialog. Steps to reproduce:

1. A valid workbasket (A) is saved in the admin.

2. The key for workbasket A is removed from the LDAP.

3. When calling workbasket A, an unclear error message “Unknown Error occurred” appears. If any changes are then attempted to be saved in the affected workbasket A, an error message “Workbasket cannot be saved” occurs, and the change is not saved in the database.

4. If a switch is made to another error-free workbasket B, no changes can be saved there either. Additionally, there is no longer any dialog asking whether the change was saved or not.

Make the error message “Unknown Error occurred” be more specific, e.g. (AccessKey/groupCN “ABC” not found in LDAP). 
Furthermore, after the error occurs, changes to other workbaskets should still be possible, and dialogs like “Workbasket with key xxx was updated” should continue to appear.

Current workaround: Logging out and then in for the admin

bruderj15:
## Definition of Done

  - will be squashed on merge

### DIFF
--- a/rest/kadai-rest-spring/src/main/java/io/kadai/common/rest/ldap/LdapClient.java
+++ b/rest/kadai-rest-spring/src/main/java/io/kadai/common/rest/ldap/LdapClient.java
@@ -48,6 +48,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.core.env.Environment;
+import org.springframework.ldap.NameNotFoundException;
 import org.springframework.ldap.core.DirContextOperations;
 import org.springframework.ldap.core.LdapTemplate;
 import org.springframework.ldap.core.support.AbstractContextMapper;
@@ -60,9 +61,7 @@ import org.springframework.ldap.filter.WhitespaceWildcardsFilter;
 import org.springframework.ldap.support.LdapNameBuilder;
 import org.springframework.stereotype.Component;
 
-/**
- * Class for Ldap access.
- */
+/** Class for Ldap access. */
 @Component
 public class LdapClient {
 
@@ -94,7 +93,7 @@ public class LdapClient {
    * @return a list of AccessIdResources sorted by AccessId and limited to
    *     maxNumberOfReturnedAccessIds
    * @throws InvalidArgumentException if input is shorter than minSearchForLength
-   * @throws InvalidNameException thrown if name is not a valid dn
+   * @throws InvalidNameException if name is not a valid dn
    */
   public List<AccessIdRepresentationModel> searchUsersAndGroupsAndPermissions(final String name)
       throws InvalidArgumentException, InvalidNameException {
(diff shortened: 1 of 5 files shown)


bruderj15:
SonarQube stupid (due to formatting)

bruderj15:
For now this fixes the backend-side of the bug: returns an empty list of access-ids if dn does not exist
The frontend now needs to adapt accordingly.

I started the frontend with the new changes and removed user `user-1-3` from `kadai-rest-spring-example-common/.../kadai-example.ldif`, then built and started the `kadai-rest-spring-example-boot` in a **production**-like way (see [docs](https://kadai-io.github.io/kadai-doc/docs/user-guide/getting-started/exampleSpringBoot#step-14-clone-build--bundle-the-frontend)).

When now clicking on a workbasket (e.g. `basxet0`) with (still there set) owner `user-1-3` , then adding a new (still valid) user (e.g. `user-1-2`) instead and saving, works as expected.

If I try to save the workbasket immediately, with still the old `user-1-3` set as owner I get error toast: 
`The Workbasket cannot be saved since the Workbasket Information contains invalid values` and a highlight on the owner-field.
This is right so far.
However if I now try to change the owner, saving has no effect and no toasts are displayed anymore at all.

This is exactly the behavior specified in the bug-report (#1222).

web/src/app/workplace/components/task-details/task-details.component.html
```diff
@@ -62,7 +62,7 @@
             @if (currentId != 'new-task') {
               <button (click)="deleteTask()"
                 class="task-details__button--secondary"
-                mat-menu-item matTooltip="Save Task">
+                mat-menu-item matTooltip="Delete Task">
```
CRoberto1926, line 65: This has nothing to do with the bug, just fixing a wrong tooltip :)

rest/kadai-rest-spring/src/main/java/io/kadai/common/rest/ldap/LdapClient.java
```diff
@@ -103,9 +102,13 @@ public List<AccessIdRepresentationModel> searchUsersAndGroupsAndPermissions(fina
...
-      AccessIdRepresentationModel groupByDn = searchAccessIdByDn(name);
-      if (groupByDn != null) {
-        accessIds.add(groupByDn);
+      try {
+        AccessIdRepresentationModel groupByDn = searchAccessIdByDn(name);
+        if (groupByDn != null) {
+          accessIds.add(groupByDn);
+        }
+      } catch (NameNotFoundException ignore) {
+        // LDAP-DN doesn't exist => yields empty result for this search
```
CRoberto1926, line 111: How about debug logging here? I guess it may be helpful

CRoberto1926 (approved):

PR #1282, opened 2025-08-07
Closes #1269 - Improvement of handling invalid AccessId entries

### LINKED ISSUE #1269: Validate AccessIds when loading a Workbasket
Currently, when a workbasket contains an invalid or outdated AccessId (e.g., referring to a deleted user), this is not detected during loading. Users only see a generic error upon saving, without any indication of which entry caused the failure.

**Goal**
Improve the user experience by validating AccessId entries as soon as the workbasket is loaded. Any invalid entries should be visibly marked in the UI, so users can address them before saving.

lucalimbach:
## Definition of Done

  - will be squashed on merge

### DIFF
--- a/web/src/app/administration/components/workbasket-access-items/workbasket-access-items.component.spec.ts
+++ b/web/src/app/administration/components/workbasket-access-items/workbasket-access-items.component.spec.ts
@@ -16,7 +16,7 @@
  *
  */
 
-import { ComponentFixture, TestBed, waitForAsync } from '@angular/core/testing';
+import { ComponentFixture, fakeAsync, TestBed, tick, waitForAsync } from '@angular/core/testing';
 import { WorkbasketAccessItemsComponent } from './workbasket-access-items.component';
 import { DebugElement } from '@angular/core';
 import { Actions, ofActionDispatched, provideStore, Store } from '@ngxs/store';
@@ -144,4 +144,23 @@ describe('WorkbasketAccessItemsComponent', () => {
     expect(onSaveSpy).toHaveBeenCalled();
     expect(actionDispatched).toBe(true);
   });
+
+  it('should emit accessItemsValidityChanged when accessItemsGroups status changes', fakeAsync(() => {
+    const emitSpy = jest.spyOn(component.accessItemsValidityChanged, 'emit');
+
+    component.ngOnInit();
+    fixture.detectChanges();
+    tick();
+
+    const control = component.AccessItemsForm.get('accessItemsGroups')?.get('0.accessId');
+    expect(control).toBeTruthy();
+
+    control?.setValue('');
+    control?.markAsTouched();
+    control?.updateValueAndValidity();
+
+    tick();
+
+    expect(emitSpy).toHaveBeenCalledWith(false);
+  }));
 });
--- a/web/src/app/administration/components/workbasket-details/workbasket-details.component.scss
+++ b/web/src/app/administration/components/workbasket-details/workbasket-details.component.scss
@@ -72,7 +72,7 @@
     margin-left: 8px;
   }
 
-  .access-warning {
+  &__access-item-warning {
     display: flex;
     align-items: flex-start;
     gap: 4px;
@@ -84,13 +84,13 @@
     border-radius: 4px;
   }
 
-  .access-warning__icon {
+  &__access-item-warning-icon {
     font-size: 20px;
     line-height: 1;
     margin-top: 1px;
   }
 
-  .access-warning__text {
+  &__access-item-warning-text {
     font-size: 0.7rem;
     line-height: 1.1;
   }
--- a/web/src/app/administration/components/workbasket-details/workbasket-details.component.scss
+++ b/web/src/app/administration/components/workbasket-details/workbasket-details.component.scss
@@ -71,4 +71,27 @@
   &__mat-icon {
     margin-left: 8px;
   }
+
+  .access-warning {
+    display: flex;
+    align-items: flex-start;
+    gap: 4px;
+    margin-right: 5px;
+    padding: 6px 8px;
+    background-color: #fcebea;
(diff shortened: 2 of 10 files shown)


web/src/app/shared/components/type-ahead/type-ahead.component.scss
```diff
@@ -29,16 +29,16 @@
...
-  }
-
   &__error--accessId {
     white-space: nowrap;
-    padding-top: 8px;
   }
 }
 
+::ng-deep .mat-mdc-form-field-error-wrapper,
+::ng-deep .mat-mdc-form-field-hint-wrapper {
```
CRoberto1926: Is it possible to style without ::ng-deep ?

lucalimbach: This styling doesn’t work without `::ng-deep ` in this file due to Angular’s view encapsulation. However, we could move it to `web/src/theme/_type-ahead.scss` and apply it globally from there, which would allow us to avoid using ::ng-deep.

web/src/app/administration/components/workbasket-details/workbasket-details.component.scss
```diff
@@ -71,4 +71,27 @@
   &__mat-icon {
     margin-left: 8px;
   }
+
+  .access-warning {
```
CRoberto1926: AFAIK the project uses BEM Methodology (https://en.bem.info/methodology/quick-start/), so if you define `.access-warning` like this, you are declaring a block, and therefore should not be inside `.workbasket-details`.

You probably need an Element hier, like `&__access-warning`. Same goes for `access-warning__icon` and `access-warning__text`.

I would also use `access-item-warning` or `invalid-access-item`, so that "access-item" appears in the class.

web/src/app/shared/components/type-ahead/type-ahead.component.scss
CRoberto1926: I would avoid ::ng-deep if it is possible: https://angular.dev/guide/components/styling#ng-deep

Also, by directly overriding Angular Material styles we'll have issues with Angular Material updates, I guess

web/src/app/administration/components/workbasket-details/workbasket-details.component.ts
```diff
@@ -82,6 +82,7 @@ export class WorkbasketDetailsComponent implements OnInit, OnDestroy {
   @Input() expanded: boolean;
   private store = inject(Store);
   private ngxsActions$ = inject(Actions);
+  hasInvalidAccessItems = false;
```
CRoberto1926: Naming suggestion: `areAllAccessItemsValid`, initialized as `true`.

It is usually better to use booleans as "good case", it helps readability IMHO. Then on line 153 you just need to assign `isValid` without negating it, and in the Component Template negate the usage on line 32.

bruderj15:
> Copilot encountered an error and was unable to review this pull request. You can try again by re-requesting a review.

Okay bro take a break

web/src/app/administration/components/workbasket-access-items/workbasket-access-items.component.spec.ts
```diff
@@ -16,7 +16,7 @@
  *
  */
 
-import { ComponentFixture, TestBed, waitForAsync } from '@angular/core/testing';
+import { ComponentFixture, fakeAsync, TestBed, tick, waitForAsync } from '@angular/core/testing';
```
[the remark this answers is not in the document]
bruderj15: irrelevant imo

web/src/app/administration/components/workbasket-details/workbasket-details.component.ts
```diff
@@ -148,6 +149,10 @@ export class WorkbasketDetailsComponent implements OnInit, OnDestroy {
     this.store.dispatch(new OnButtonPressed(ButtonAction.SAVE));
   }
 
+  onAccessItemsValidityChanged(isValid: boolean) {
```
bruderj15: Nitpick: 
Method-name implies implementation (usage of duality in boolean algebra). Therefore one could omit the method argument.

bruderj15 (approved):
> Add `accessItems[0].accessId = "user-15-1"; `at line 160 in `workbasket-access-items.component.ts` for testing.

The release-note in the PR-Description is outdated?
Otherwise looking good.

web/src/app/administration/components/workbasket-details/workbasket-details.component.ts
lucalimbach: The child component already determines `isValid `and emits it, so the parent simply stores the value. This avoids duplicating the validity check in the parent. I can rename the method to `handleAccessItemsValidityChanged` if you prefer.

bruderj15: > The child component already determines isValid and emits it, so the parent simply stores the value. This avoids duplicating the validity check in the parent.

Yeah that's why I'd do:
```ts
onAccessItemsValidityChanged() {
  this.areAllAccessItemsValid = !this.areAllAccessItemsValid;
}
```

lucalimbach: This isn't working correctly. The child emits the absolute validity state, so toggling in the parent will invert it incorrectly on the initial emission or repeated identical values — direct assignment keeps them in sync.
image
image

lucalimbach: But i can rename the handler to handleAccessItemsValidityChanged for clarity.

bruderj15: Alright, good catch then. Renaming good.

bruderj15 (commented):
Tests are broken now whoopsie

bruderj15 (approved):

Issue #1049, opened 2025-05-23
Automate gathering of release-notes


bruderj15:

bruderj15:
[Release Drafter](https://github.com/marketplace/actions/release-drafter) provides access to the full body of all merged PRs.
Custom postprocessing (outside of Release Drafter) could then parse and process the PR-Body to remove all noise.
Seems a little hacky though.

May just write a custom action.

bruderj15:
What about an AI-Approach?

leantios:
What is the status here? Is this done/ in progress?

bruderj15:
Will be stall for a while.

bruderj15:
Yeah, not fox the work, won't fix

PR #1291, opened 2025-08-11
Closes #1264 - Force-update tag to reflect actual release-version in poms

### LINKED ISSUE #1264: Fix release workflow: Missing git commit for release version
## Problem

The current release workflow has a gap in version control tracking. When a new release is created (e.g., v1.2.3), the workflow correctly
builds and deploys artifacts with the release version, but never commits the code with that release version to git.

Therefore, the assets of the latest release [v10.1.0](https://github.com/kadai-io/kadai/releases/tag/v10.1.0) includes the SNAPSHOT version in the pom.xml files instead of the actual release version.

## Current Behavior

1. Tag v1.2.3 is pushed (triggering the release)
2. Workflow sets project version to 1.2.3 (via `ci/change_version.sh -m .`)
3. Artifacts are built and released with version 1.2.3
4. Workflow increments version to 1.2.4-SNAPSHOT and commits this to git
5. Result: No git commit exists with the actual release version 1.2.3

## Expected Behavior

The workflow should:
1. Set project version to release version (1.2.3)
2. Commit this release version to git
3. Ensure the release tag points to this commit
4. Then increment to next snapshot version (1.2.4-SNAPSHOT) and commit

## Impact

- Makes it difficult to track exactly what code was released
- Git history is incomplete - shows only snapshot versions
- Complicates debugging issues in specific releases
- Violates typical Git workflow best practices

## Files Involved

- .github/workflows/continuous-integration.yml (lines 489-512, update_version job)
- ci/change_version.sh

## Proposed Solution

Modify the update_version job to:
1. First commit the release version before incrementing
2. Ensure the release tag points to the correct commit
3. Then commit the snapshot version

## Disclaimer

Issue text was created with Claude Code.

bruderj15:
## Definition of Done

  - will be squashed on merge


.github/workflows/continuous-integration.yml
```diff
@@ -464,8 +464,19 @@ jobs:
...
-        run: ci/change_version.sh -m .
+      - name: Change versions to match tag and update it
+        run: |
+          git config --global user.email "$GH_EMAIL"
+          git config --global user.name "$GH_USERNAME"
+          ci/change_version.sh -m .
+          git add "./*pom.xml"
+          git commit -m "Updated poms to version ${GITHUB_REF##refs/tags/v}"
+          git tag -f "${GITHUB_REF##refs/tags/}"
+          git push --force-with-lease origin "${GITHUB_REF##refs/tags/}"
```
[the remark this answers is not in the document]
bruderj15: The sole intent is to overwrite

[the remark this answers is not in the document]
bruderj15: @copilot will this re-trigger the CI?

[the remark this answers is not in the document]
bruderj15: I cannot see how this would be any more risky. Declined.

.github/release-drafter.yml
leantios, line 1: Why did you remove this file?

.github/workflows/continuous-integration.yml
```diff
@@ -464,8 +466,19 @@ jobs:
...
         env:
           GPG_KEY: ${{ secrets.GPG_KEY }}
-      - name: Change versions to match tag
-        run: ci/change_version.sh -m .
+      - name: Change versions to match tag and update it
+        run: |
+          git config --global user.email "$GH_EMAIL"
+          git config --global user.name "$GH_USERNAME"
+          ci/change_version.sh -m .
+          git add "./*pom.xml"
```
lucalimbach: git add ‘./*pom.xml’ seems to only affect the root file pom.xml. Shouldn't the module POMs also be included?

.github/release-drafter.yml
bruderj15: Was intended for #1049 and there for POC only. We're not going this route anymore, therefore removed it.

.github/workflows/continuous-integration.yml
```diff
@@ -464,8 +466,19 @@ jobs:
...
-        run: ci/change_version.sh -m .
+      - name: Change versions to match tag and update it
+        run: |
+          git config --global user.email "$GH_EMAIL"
+          git config --global user.name "$GH_USERNAME"
+          ci/change_version.sh -m .
+          git add .
+          git commit -m "Updated poms to version ${GITHUB_REF##refs/tags/v}"
+          git tag -f "${GITHUB_REF##refs/tags/}"
+          git push --force-with-lease origin "${GITHUB_REF##refs/tags/}"
```
lucalimbach, line 477: The requirements mention committing both the release version and next snapshot version, but the current implementation only pushes the tag. Should these commits also be pushed to master to complete the expected release flow, or is the tag-only approach intentional?

bruderj15: It was intentional, but thinking about it again: The commit of the tag should be on main too, I will add it.

.github/workflows/continuous-integration.yml
```diff
@@ -464,8 +466,20 @@ jobs:
...
+          git push origin main
```
lucalimbach, line 478: The workflow tags and commits the release version, but doesn’t bump to the next -SNAPSHOT.
Should we add a step (like in the update_version job) so the main branch immediately continues with a snapshot version after a release?

bruderj15: Why should it bump to the next snapshot here?
As you mentioned, this is already done in the following job.
May you elaborate?

lucalimbach (dismissed):
looks good!

.github/workflows/continuous-integration.yml
leantios: What happens now if the pipeline runs for a commit that is not the newest commit on main? 
Scenario:
Someone commits A to main
Commits B and C are added to main
A tag is created pointing to commit A (the older commit)
Pipeline triggers from that tag
The push to main could be an issue, no?

.github/workflows/continuous-integration.yml
```diff
@@ -464,8 +466,20 @@ jobs:
...
+        env:
```
leantios, line 478: main is called master

leantios (changes requested):

.github/workflows/continuous-integration.yml
bruderj15: whoopsie

.github/workflows/continuous-integration.yml
bruderj15: Interesting thought.
Your example is inevitable. It already is with the current state where we push the next snapshot version.
I don't think it's too much of a (real) problem.

However you point to an actual problem: Tagging from another branch - think patch-release for an older version (like we did with 9.3.2).
Then we would push the wrong actual _new_ version to main. That'd be bad.

Makes me think if we really want to address the point @lucalimbach mentioned.
I'd propose to not push to main then - only updating the version for the tag.

leantios: You got a point there. I think there is no way to push a commit right after another commit in the same (unknown) branch?

bruderj15: Nope. History is append-only.

leantios (approved):

PR #1345, opened 2025-09-05
Closes #1226: New Endpoint for Update multiple Tasks

### LINKED ISSUE #1226: New Endpoint for Update multiple Tasks
New REST-API:
 - Endpoint /api/v1/tasks/bulkupdate
 - Method PATCH
 - Request Body Parameters:
   - List TaskIDs
   - List of Values to change
   - Use new data structure, not task data structure
 - compare with existing APIs
 - check which attributes can be updated and include all

New JAVA-API
Authority Check
Error Handling:
 - One Update Statement in Database, not possible to show each errors
 - If no permission for one of the tasks, all other tasks are still updated, and an individual error is given to the failed task

New Mapping in MyBatis for the BulkUpdate
Unit and Integration Tests
Documentation

jamesrdi:
## Definition of Done

  - will be squashed on merge
  - New API Endpoint for bulk update of tasks: /api/v1/tasks/bulkupdate
  - Request Body for Bulk Update of tasks can be found in the [Swagger Doc](https://kadai-io.azurewebsites.net/kadai/swagger-ui/index.html#/task-controller) under bulkUpdateTasks endpoint
  - createTask, updateTask and bulkUpdateTask now throws ServiceLevelViolationException instead of InvalidArgumentException when service level specified in request body is violated and enforce service level is true, ServiceLevelViolation will return error code 422_UNPROCESSABLE_ENTITY


lib/kadai-core-test/src/test/java/acceptance/ArchitectureTest.java
```diff
@@ -435,7 +436,7 @@ void packagesShouldBeFreeOfCyclicDependencies() {
   }
 
   @Test
-  //@Disabled("this has way too many false positives during development without refactoring")
+  @Disabled("this has way too many false positives during development without refactoring")
```
bruderj15: It used to work. You introduced a cycle then? Which one?
(I think any cycle can be avoided)

lib/kadai-core/src/main/java/io/kadai/task/api/models/TaskPatch.java
```diff
@@ -0,0 +1,842 @@
...
+
+import io.kadai.classification.api.models.ClassificationSummary;
+import io.kadai.task.api.CallbackState;
+import io.kadai.task.api.TaskState;
+import io.kadai.workbasket.api.models.WorkbasketSummary;
+import java.time.Instant;
+import java.util.List;
+import java.util.Map;
+
+public interface TaskPatch {
```
bruderj15: By its nature TaskPatch is a command (think CQRS).
A command is read-only because instructions are immutable.
Think: As a task-owner of task x I want to change due-date and received-date. That's an immutable statement.
Therefore I wonder about the setters in the interface here.
They should not be here.

lib/kadai-core/src/main/java/io/kadai/task/api/models/TaskPatch.java
```diff
@@ -0,0 +1,842 @@
...
+   * @return taskId
+   */
+  String getId();
+
+  /**
+   * Sets the id of the {@linkplain TaskPatch}.
+   *
+   * @param id new id
+   */
+  void setId(String id);
```
bruderj15: TLDR: The type `TaskPatch` should only contain fields the user can actually change.

Including the identifier in `TaskPatch` (basically a command) is hurting separation of concerns.
The command no longer only expresses the user-intent (what) but also includes the target (where).
Especially in the context of a bulk-operation this is unfortunate because it does not align with the user intention:

_As a user I want to update the due-date for tasks x, y and z in bulk_.

With the current approach we would need a `TaskPatch` for every target.
This is contrary to the user-goal.

lib/kadai-core/src/main/java/io/kadai/task/api/TaskService.java
```diff
@@ -142,6 +144,8 @@ public interface TaskService {
    * @throws ObjectReferencePersistenceException if an {@linkplain ObjectReference} with the same
    *     {@linkplain ObjectReference#getId() id} was added to the {@linkplain Task} multiple times
    *     without using {@linkplain Task#addSecondaryObjectReference(ObjectReference)}
+   * @throws ServiceLevelViolationException if the {@linkplain Task#getDue() due} and {@linkplain
+   *     Task#getPlanned() planned} do not match service level
    */
   Task createTask(Task taskToCreate)
```
bruderj15, line 149: Changing from InvalidArgumentException to ServiceLevelException is a breaking-change for Java- and REST-Api. Make sure to thoroughly document it in the release-notes.

lib/kadai-core/src/main/java/io/kadai/task/api/TaskService.java
```diff
@@ -1631,4 +1639,18 @@ ObjectReference newObjectReference(
    * @return a {@linkplain TaskCommentQuery}
    */
   TaskCommentQuery createTaskCommentQuery();
+
+  /**
+   * Bulk updates multiple tasks by their IDs.
+   *
+   * <p>Updates only the specified fields for each task. If the user lacks permission for a task,
+   * that task is skipped and an error is reported for it, but other tasks are still updated. The
```
bruderj15: `skipped` may mislead.
In general: skip == discard/swallow/silent failure

lib/kadai-core/src/main/java/io/kadai/task/api/TaskService.java
```diff
@@ -1631,4 +1639,18 @@ ObjectReference newObjectReference(
...
+   * update is performed in a single database statement for all permitted tasks.
```
bruderj15: Amount of statements should not be leaked here (also provides no value).
You probably meant transaction?
Doesn't quite align with the partial-failure approach from the previous sentence.

lib/kadai-core/src/main/java/io/kadai/task/internal/models/TaskPatchImpl.java
```diff
@@ -0,0 +1,772 @@
...
+  }
+
+  /**
+   * Applies all non-null fields from this TaskPatch to the provided TaskImpl. Only fields that are
+   * not null will be written, preserving existing values otherwise.
+   *
+   * @param task the {@link TaskImpl} to update, may be {@code null}
+   * @return the updated {@link TaskImpl}, or {@code null} if the input was {@code null}
+   */
+  public TaskImpl toTaskImpl(TaskImpl task) {
```
bruderj15: There exists something similar somewhere, right?
But it's for `TaskImpl -> TaskImpl -> TaskImpl`...?

lib/kadai-core/src/main/java/io/kadai/task/internal/models/TaskPatchImpl.java
```diff
@@ -0,0 +1,772 @@
...
+    if (task == null) {
+      return null;
+    }
+
+    // Base/scalar fields
+    if (id != null) {
+      task.setId(id);
+    }
```
bruderj15: Suspicious and related to a previous remark of mine.
Id shall not be changed.

lib/kadai-core/src/main/java/io/kadai/task/internal/TaskServiceImpl.java
```diff
@@ -2589,4 +2596,68 @@ private boolean checkEditTasksPerm(TaskSummary task) {
...
+
+    String userId = kadaiEngine.getEngine().getCurrentUserContext().getUserid();
+    try {
+      kadaiEngine.openConnection();
+
+      for (String taskId : taskIds) {
+        try {
+          TaskImpl oldTaskImpl = (TaskImpl) getTask(taskId);
+          if (!checkEditTasksPerm(oldTaskImpl)) {
+            throw new NotAuthorizedOnWorkbasketException(
```
bruderj15: Throw-catch within the same block is an anti-pattern.
Should instead directly:
```java
bulkLog.addError(taskId, new NotAuthorizedOnWorkbasketException(..));
```

rest/kadai-rest-spring/src/main/java/io/kadai/task/rest/assembler/TaskRepresentationModelAssembler.java
```diff
@@ -232,4 +235,201 @@ private void verifyCorrectCustomAttributesFormat(TaskRepresentationModel repMode
               + "\"customAttributes\": [{\"key\": \"someKey\",\"value\": \"someValue\"},{...}])");
     }
   }
+
+  public TaskPatchImpl toPatchModel(TaskPatchRepresentationModel repModel) {
```
bruderj15: naming seems weird as we're converting `repModel` `toPatchModel` and result in something without `Model` suffix

rest/kadai-rest-spring/src/main/java/io/kadai/task/rest/models/TaskPatchRepresentationModel.java
```diff
@@ -0,0 +1,154 @@
...
+  protected Integer numberOfComments;
+
+  @Schema(name = "isRead", description = "Indicator if the task has been read.")
+  protected Boolean isRead;
+
+  @Schema(name = "isTransferred", description = "Indicator if the task has been transferred.")
+  protected Boolean isTransferred;
+
+  @Schema(name = "isReopened", description = "Indicator if the task has been reopened.")
+  protected Boolean isReopened;
```
bruderj15: These fields are implied by internal invariants and therefore read-only

rest/kadai-rest-spring/src/main/java/io/kadai/task/rest/TaskApi.java
```diff
@@ -141,9 +145,15 @@ public interface TaskApi {
             }),
         @ApiResponse(
             responseCode = "400",
-            description = "INVALID_ARGUMENT",
+            description = "INVALID_ARGUMENT, SERVICE_LEVEL_VIOLATION",
```
bruderj15: `INVALID_ARGUMENT` is not required anymore, right?

rest/kadai-rest-spring/src/main/java/io/kadai/task/rest/TaskApi.java
```diff
@@ -141,9 +145,15 @@ public interface TaskApi {
             }),
         @ApiResponse(
             responseCode = "400",
```
bruderj15, line 147: This should've always been a 422 imo.
It certainly is now.

bruderj15 (commented):
Inline-JSON blows up the release-notes. I'd prefer a link to Swagger-Docs.

lib/kadai-core-test/src/test/java/acceptance/ArchitectureTest.java
jamesrdi: I forgot to comment it out again, I enabled the "Disabled" for debugging purpose. But actually there is no more cycle.

lib/kadai-core/src/main/java/io/kadai/task/api/models/TaskPatch.java
jamesrdi: Ok, then I would delete the setters and use builders instead.

bruderj15: Why that?
The setters can still be a part of `TaskPatchImpl`.

jamesrdi: Ah I see what you mean, ok then I will leave the setters as it is in TaskPatchImpl

rest/kadai-rest-spring/src/main/java/io/kadai/task/rest/TaskApi.java
jamesrdi: We still need it, the effected endpoints still use "INVALID_ARGUMENT", for example in createTask, resolveWorkbasket() in TaskServiceImpl throws "INVALID_ARGUMENT"

lib/kadai-core/src/main/java/io/kadai/task/api/TaskService.java
jamesrdi: You're right, it is supposed to be transaction

lib/kadai-core/src/main/java/io/kadai/task/internal/models/TaskPatchImpl.java
jamesrdi: I am not quite sure, but if it exists, how can it help? We need to create a new TaskImpl that applies all non-null fields from the TaskPatch to the old TaskImpl

lib/kadai-core/src/main/java/io/kadai/task/api/TaskService.java
bruderj15: If it was all done in one transaction, how could we do partial failure?

jamesrdi: You’re right — I mixed this up and assumed that opening a connection meant we were in a single transaction. Thanks for pointing that out.

rest/kadai-rest-spring/src/main/java/io/kadai/task/rest/models/TaskPatchRepresentationModel.java
jamesrdi: The fields numberOfComments, isRead, isTransferred and isReopened are indeed updatable, I have tried it with the single update API

rest/kadai-rest-spring/src/main/java/io/kadai/task/rest/TaskApi.java
jamesrdi: Yes I agree, Service Level Violation should be with 422. Now that you mention it, in the doc, invalid argument has status code 400, but how it is implemented in KadaiRestExceptionHandler.java, it actually throws 500. We can later open a separate issue to fix the invalid argument exception

rest/kadai-rest-spring/src/main/java/io/kadai/task/rest/models/TaskPatchRepresentationModel.java
bruderj15: Well, then `TaskService::updateTask` does it wrong as well.
I'll reach out to the PO.

rest/kadai-rest-spring/src/main/java/io/kadai/task/rest/TaskApi.java
```diff
@@ -180,7 +184,14 @@ public interface TaskApi {
                             AttachmentPersistenceException.class,
                             ObjectReferencePersistenceException.class
                           }))
-            })
+            }),
+          @ApiResponse(
```
bruderj15: These breaking changes also need to be documented in the PR-Description

rest/kadai-rest-spring/src/main/java/io/kadai/task/rest/TaskController.java
```diff
@@ -531,6 +535,38 @@ public ResponseEntity<TaskRepresentationModel> updateTask(
...
+      throw new InvalidArgumentException("fieldsToUpdate must not be empty");
+    }
+
+    if (requestModel.getFieldsToUpdate().getOwner() != null) {
+      throw new InvalidArgumentException("Owner can only be updated by transferring tasks");
+    }
+    if (requestModel.getFieldsToUpdate().getAttachments() != null) {
+      throw new InvalidArgumentException(
+          "Attachments cannot be updated in bulk as they are task " + "specific");
+    }
```
bruderj15: Why are these part of the type then?
Might as well just remove them.

rest/kadai-rest-spring/src/main/java/io/kadai/task/rest/TaskController.java
```diff
@@ -531,6 +535,38 @@ public ResponseEntity<TaskRepresentationModel> updateTask(
...
+  @Transactional(rollbackFor = Exception.class)
+  public ResponseEntity<BulkOperationResultsRepresentationModel> bulkUpdateTasks(
+      @RequestBody TaskBulkUpdateRepresentationModel requestModel) {
+
+    if (requestModel.getTaskIds() == null || requestModel.getTaskIds().isEmpty()) {
+      throw new InvalidArgumentException("taskIds must not be empty");
+    }
+    if (requestModel.getFieldsToUpdate() == null) {
+      throw new InvalidArgumentException("fieldsToUpdate must not be empty");
+    }
```
bruderj15, line 551: Every Dev has two choices for empty `PATCH`:
- fail (400 or 422)
- succeed as no-op (200)

The latter is much more client-friendly and in general recommended.

bruderj15 (commented):
What about this?

https://github.com/kadai-io/kadai/blob/0b14c7012aa3ec879dc79d31c727eadaf7abd6d3/lib/kadai-core/src/main/java/io/kadai/task/api/TaskService.java#L1296-L1297

I think we may want to deprecate it in favor of the - here introduced - more general API.

rest/kadai-rest-spring/src/main/java/io/kadai/task/rest/TaskApi.java
bruderj15: Outsourced to #1359

rest/kadai-rest-spring/src/main/java/io/kadai/task/rest/models/TaskPatchRepresentationModel.java
bruderj15: Consultation with PO lead to:
- Fields `numberOfComments`, `isTransferred` and `isReopened` shall not be updateable by the user.
- Field `isRead` shall remain updateable.

rest/kadai-rest-spring/src/main/java/io/kadai/task/rest/TaskApi.java
jamesrdi: I have updated the release notes again, do tell me if I need to add something more

rest/kadai-rest-spring/src/main/java/io/kadai/task/rest/TaskController.java
jamesrdi: At first I thought the patch model could be used for purposes other than just bulkUpdateTasks, which might require Attachments or Owner to be updated, so we could reuse the patch model. But you're right, it may be better for now to just remove them and maybe add them again later if they are needed in other use cases.

lib/kadai-core/src/main/java/io/kadai/task/api/models/TaskPatch.java
```diff
@@ -0,0 +1,427 @@
...
+   * @return the Tasks isTransferred flag
+   */
+  Boolean isTransferred();
+
+  /**
+   * Returns whether the {@linkplain Task} has been previously reopened.
+   *
+   * @return true, if Tasks has been reopened before
+   */
+  Boolean isReopened();
```
bruderj15: these should go

lib/kadai-core/src/main/java/io/kadai/task/api/models/TaskPatch.java
```diff
@@ -0,0 +1,427 @@
...
+   * @return the manualPriority of the TaskPatch
+   */
+  Integer getManualPriority();
+
+  /**
+   * Returns the count of the comments of the {@linkplain TaskPatch}.
+   *
+   * @return numberOfComments
+   */
+  Integer getNumberOfComments();
```
bruderj15: away with this peasant as well!

lib/kadai-core/src/main/java/io/kadai/task/api/models/TaskPatch.java
```diff
@@ -0,0 +1,427 @@
...
+   * @return the Tasks primary {@linkplain ObjectReference}
+   */
+  ObjectReference getPrimaryObjRef();
+
+  /**
+   * Returns the priority of the {@linkplain Task}.
+   *
+   * @return priority
+   */
+  Integer getPriority();
```
bruderj15: This doesn't belong in here as well, does it?

The priority itself is read-only.
If one wants to update it manual, one has shall use `manualPriority`

lib/kadai-core/src/main/java/io/kadai/task/api/models/TaskPatch.java
```diff
@@ -0,0 +1,427 @@
...
+   * @return owner
+   */
+  String getOwner();
+
+  /**
+   * Returns long name of the owner of the {@linkplain TaskPatch}.
+   *
+   * @return the long name of the owner
+   */
+  String getOwnerLongName();
```
bruderj15: i doubt this shall be changeable

bruderj15 (commented):
I stopped after a few comments.
Please thoroughly revisit the interface `TaskPatch`.

It contains multiple fields that shall not be updateable.
I only added comments for a few.

lib/kadai-core/src/main/java/io/kadai/task/api/models/TaskPatch.java
```diff
@@ -0,0 +1,376 @@
...
+   * @return the claimed Instant
+   */
+  Instant getClaimed();
+
+  /**
+   * Returns the modified TimeStamp of the {@linkplain TaskPatch}.
+   *
+   * @return the last modified Instant
+   */
+  Instant getModified();
```
bruderj15: I don't think that these timestamps belong in here

lib/kadai-core/src/main/java/io/kadai/task/api/models/TaskPatch.java
```diff
@@ -0,0 +1,376 @@
...
+   * @return the due Instant
+   */
+  Instant getDue();
+
+  /**
+   * Returns the completed TimeStamp of the {@linkplain TaskPatch}.
+   *
+   * @return the completed Instant
+   */
+  Instant getCompleted();
```
bruderj15: neither does this

bruderj15 (commented):
I think there are still fields in `TaskPatch` that are not eligible for change - e.g. most of the timestamps.

You can simply compare what the interfaces `Task` and `TaskService` offer.

NB: I am not sure if we need an interface for `TaskPatch`. The type does not directly describe anything domain-related. It's simply data.
Maybe ask yourself the question: _Does anyone benefit from the abstraction?_
I think no.
The user never has it in co-variant position (= consuming - read only).
He only uses it contra-variant.
Consider simply using a Java-Record.

bruderj15:
Also, why don't I see the CI being run?

jamesrdi:
> I think there are still fields in `TaskPatch` that are not eligible for change - e.g. most of the timestamps.
> 
> You can simply compare what the interfaces `Task` and `TaskService` offer.
> 
> NB: I am not sure if we need an interface for `TaskPatch`. The type does not directly describe anything domain-related. It's simply data. Maybe ask yourself the question: _Does anyone benefit from the abstraction?_ I think no. The user never has it in co-variant position (= consuming - read only). He only uses it contra-variant. Consider simply using a Java-Record.


Unfortunately we cannot implement TaskPatch as a Record, because we have a PojoTest that has `GetterMustExistRule()` in the test `validateGetters`

bruderj15:
> Unfortunately we cannot implement TaskPatch as a Record, because we have a PojoTest that has `GetterMustExistRule()` in the test `validateGetters`

This test isn't universal law.
You can adjust it :)

jamesrdi:
> > Unfortunately we cannot implement TaskPatch as a Record, because we have a PojoTest that has `GetterMustExistRule()` in the test `validateGetters`
> 
> This test isn't universal law. You can adjust it :)

I see. Ok then I'll adjust it

lib/kadai-core-test/src/test/java/acceptance/PojoTest.java
```diff
@@ -56,6 +56,7 @@ class PojoTest {
                           && !javaClass.getSimpleName().equals("LongItem")
                           // This is a record, it has a getter per definition
                           && !javaClass.getSimpleName().equals("DurationPrioHolder")
+                          && !javaClass.getSimpleName().equals("TaskPatch")
                           // This is a record, it has a getter per definition
                           && !javaClass.getSimpleName().equals("CustomHoliday"))
```
bruderj15: there's probably are more generic way for this.
I imagine something like `Type::isRecord`.

lib/kadai-core/src/main/java/io/kadai/task/api/TaskService.java
```diff
@@ -1292,7 +1299,9 @@ List<String> updateTasks(
    * @return a list of the {@linkplain Task#getId() ids} of all modified {@linkplain Task Tasks}
    * @throws InvalidArgumentException if the given customFieldsToUpdate are NULL or empty
    * @see #updateTasks(ObjectReference, Map)
+   * @deprecated Use {@link #bulkUpdateTasks(List, TaskPatch)} instead.
    */
+  @Deprecated
```
bruderj15: I'd add `forRemoval = "12.0.0"`.
(for all occurences)

lib/kadai-core/src/main/java/io/kadai/task/internal/TaskServiceImpl.java
```diff
@@ -2589,4 +2595,209 @@ private boolean checkEditTasksPerm(TaskSummary task) {
...
+      kadaiEngine.openConnection();
+
+      for (String taskId : taskIds) {
+        try {
+          TaskImpl oldTaskImpl = (TaskImpl) getTask(taskId);
+          if (!checkEditTasksPerm(oldTaskImpl)) {
+            bulkLog.addError(
+                taskId,
+                new NotAuthorizedOnWorkbasketException(
+                    kadaiEngine.getEngine().getCurrentUserContext().getUserid(),
```
bruderj15: can reuse the one living from L2604 :)

lib/kadai-core/src/main/java/io/kadai/task/internal/TaskServiceImpl.java
```diff
@@ -2589,4 +2595,209 @@ private boolean checkEditTasksPerm(TaskSummary task) {
         query.idIn(workbasketId).callerHasPermissions(WorkbasketPermission.EDITTASKS).single();
     return workbasket != null;
   }
+
+  @Override
+  public BulkOperationResults<String, KadaiException> bulkUpdateTasks(
```
bruderj15, line 2601: I do wonder a little here.
I thought we were simply going to reuse `TaskService::updateTask`?

But right now we're essentially duplicating logic.

rest/kadai-rest-spring/src/main/java/io/kadai/task/rest/assembler/TaskRepresentationModelAssembler.java
```diff
@@ -232,4 +236,154 @@ private void verifyCorrectCustomAttributesFormat(TaskRepresentationModel repMode
...
   }
+
+  public TaskPatch toPatch(TaskPatchRepresentationModel repModel) {
+    if (repModel.getCustomAttributes() != null) {
+      verifyCorrectCustomAttributesFormat(repModel.getCustomAttributes());
+    }
+
+    TaskPatchBuilder builder = new TaskPatchBuilder();
+
+    if (repModel.getReceived() != null) {
```
bruderj15: I don't think the null-guards are necessary, aren't they?
All fields of `TaskPatch` are nullable.

rest/kadai-rest-spring/src/main/java/io/kadai/task/rest/models/BaseTaskRepresentationModel.java
```diff
@@ -0,0 +1,471 @@
...
+import io.kadai.workbasket.rest.models.WorkbasketSummaryRepresentationModel;
+import io.swagger.v3.oas.annotations.media.Schema;
+import jakarta.validation.constraints.NotNull;
+import java.time.Instant;
+import java.util.ArrayList;
+import java.util.List;
+import org.springframework.hateoas.RepresentationModel;
+
+@Schema(description = "This class contains the nullable fields of Task.")
+public class BaseTaskRepresentationModel
```
bruderj15: I do see why you want to introduce this shared parent type.
I don't think I like it though.

While it does reduce the amount of POJO-Boiler, it introduces slight coupling.
Further - although considered stupid data (DTO) - this class now has no real meaning - making it harder to grasp the bigger picture.

Another point: 
This type contains field `taskId` which _would_ seem appropriate for a shared parent type for Task-related DTOs.
However the `TaskPatchRepresentationModel` now also contains this field - which is wrong.

I would personally keep the DTOs entirely separated and pay the (little) price for the boiler.

rest/kadai-rest-spring/src/main/java/io/kadai/task/rest/TaskApi.java
```diff
@@ -1726,7 +1746,157 @@ ResponseEntity<TaskRepresentationModel> updateTask(
...
+   *     along with its values
+   * @return the taskIds and corresponding ErrorCode of tasks failed to be updated
+   */
+  @Operation(
+      summary = "Updates multiple Tasks",
+      description =
+          "This endpoint updates a list of Tasks listed in the body with the fields and values "
+              + "listed in the body, if possible. Tasks that can be updated without throwing an "
+              + "exception get updated independent of other Tasks. If the update of a Task throws "
+              + "an exception, then the Task will not get updated.",
```
bruderj15: Last two sentences can simply be summarized as: Fails partially.

rest/kadai-rest-spring/src/main/java/io/kadai/task/rest/TaskController.java
```diff
@@ -531,6 +535,30 @@ public ResponseEntity<TaskRepresentationModel> updateTask(
...
 
+  @PatchMapping(path = RestEndpoints.URL_TASKS_BULK_UPDATE)
+  @Transactional(rollbackFor = Exception.class)
+  public ResponseEntity<BulkOperationResultsRepresentationModel> bulkUpdateTasks(
+      @RequestBody TaskBulkUpdateRepresentationModel requestModel) {
+
+    if (requestModel.getTaskIds() == null || requestModel.getTaskIds().isEmpty()) {
+      throw new InvalidArgumentException("taskIds must not be empty");
+    }
+    if (requestModel.getFieldsToUpdate() == null) {
```
bruderj15, line 546:
```suggestion
    if all fields of TaskPatch are null {
```

All fields null could happen.
If the object itself is null, it's highly likely to be a client error (400).

rest/kadai-rest-spring/src/main/java/io/kadai/task/rest/TaskController.java
```diff
@@ -531,6 +535,30 @@ public ResponseEntity<TaskRepresentationModel> updateTask(
...
   }
 
+  @PatchMapping(path = RestEndpoints.URL_TASKS_BULK_UPDATE)
+  @Transactional(rollbackFor = Exception.class)
+  public ResponseEntity<BulkOperationResultsRepresentationModel> bulkUpdateTasks(
+      @RequestBody TaskBulkUpdateRepresentationModel requestModel) {
+
+    if (requestModel.getTaskIds() == null || requestModel.getTaskIds().isEmpty()) {
+      throw new InvalidArgumentException("taskIds must not be empty");
+    }
```
bruderj15: consider doing it equally to the two lines below

bruderj15 (commented):
Make sure to add a case for the newly introduced exception here:
https://github.com/kadai-io/kadai/blob/d43a4d836536cca3df7d7ac00a49494d30e8a564/rest/kadai-rest-spring/src/test/java/io/kadai/common/rest/ExceptionErrorKeyTest.java#L61

lib/kadai-core-test/src/test/java/acceptance/PojoTest.java
jamesrdi: Maybe, I was just following the DurationPrioHolder class. But I can look into it

lib/kadai-core/src/main/java/io/kadai/task/internal/TaskServiceImpl.java
jamesrdi: Sorry, What do you mean?

lib/kadai-core/src/main/java/io/kadai/task/internal/TaskServiceImpl.java
jamesrdi: Actually yes you're correct. The reason behind that previously was that updateTask() checks for concurrency, where as in bulkUpdate we do not want to check for concurrency. But now since we removed the fields unchangeable by the user in the TaskPatch (e.g modified) and we always populate the newTask with the timestamps of the oldTask fetched from the database, causing the concurrency check to always pass, we can now call updateTask() from bulkUpdateTasks()

rest/kadai-rest-spring/src/main/java/io/kadai/task/rest/assembler/TaskRepresentationModelAssembler.java
jamesrdi: Yes you're right

rest/kadai-rest-spring/src/main/java/io/kadai/task/rest/TaskController.java
jamesrdi: What do you mean? Should we return 200 when taskIds is empty?

rest/kadai-rest-spring/src/main/java/io/kadai/task/rest/models/BaseTaskRepresentationModel.java
jamesrdi: Yes, I agree

rest/kadai-rest-spring/src/main/java/io/kadai/task/rest/TaskController.java
bruderj15: Yeah I'd do it that way. Much more client-friendly.

lib/kadai-core/src/main/java/io/kadai/task/internal/TaskServiceImpl.java
```diff
@@ -963,6 +968,7 @@ public BulkOperationResults<String, KadaiException> forceCompleteTasks(List<Stri
   }
 
   @Override
+  @Deprecated(forRemoval = true, since = "12.0.0")
```
bruderj15:
```suggestion
  @Deprecated(forRemoval = true, since = "11.0.0")
```

I told you wrong. This is right now :)

rest/kadai-rest-spring/src/main/java/io/kadai/task/rest/models/TaskPatchRepresentationModel.java
```diff
@@ -0,0 +1,499 @@
...
+import io.swagger.v3.oas.annotations.media.Schema;
+import java.time.Instant;
+import java.util.List;
+
+@Schema(
+    description =
+        "Patch model for Task with all fields nullable - null values are ignored during updates. "
+            + "Note: Some complex fields like workbasketSummary, primaryObjRef, "
+            + "secondaryObjectReferences, and attachments are not yet supported in bulk updates and"
+            + " will be ignored.")
```
bruderj15: Is this _note_ actually true?

rest/kadai-rest-spring/src/test/java/io/kadai/task/rest/TaskControllerIntTest.java
```diff
@@ -2240,6 +2240,325 @@ void should_ThrowError_When_UpdatingTaskWithBadAttachment() {
...
+              .retrieve()
+              .toEntity(TaskRepresentationModel.class);
+
+      assertThat(taskResponse.getBody()).isNotNull();
+      assertThat(taskResponse.getBody().getName()).isEqualTo("Mixed Scenario Test");
+      assertThat(taskResponse.getBody().getDescription())
+          .isEqualTo("Testing various failure scenarios");
+    }
+  }
+
```
bruderj15, line 2648: Please also add tests for the edge-cases of empty payloads.

bruderj15 (commented):
We're getting there! :)

lib/kadai-core/src/main/java/io/kadai/task/internal/TaskServiceImpl.java
bruderj15: The user-id already has an rvalue in line 2604

rest/kadai-rest-spring/src/main/java/io/kadai/task/rest/models/TaskPatchRepresentationModel.java
jamesrdi: You're right, it's not. I will fix it

rest/kadai-rest-spring/src/test/java/io/kadai/task/rest/TaskControllerIntTest.java
jamesrdi: You're right I will add it.

lib/kadai-core-test/src/test/java/acceptance/PojoTest.java
```diff
@@ -57,7 +57,8 @@ class PojoTest {
                           // This is a record, it has a getter per definition
                           && !javaClass.getSimpleName().equals("DurationPrioHolder")
                           // This is a record, it has a getter per definition
-                          && !javaClass.getSimpleName().equals("CustomHoliday"))
+                          && !javaClass.getSimpleName().equals("CustomHoliday")
+                          && !javaClass.getSimpleName().equals("TaskPatch"))
```
jamesrdi, line 61: I have tried to filter instead for only classes that are not Records, but somehow in the pipeline it failed to filter out the classes. I would say for now we stick to this

lib/kadai-core/src/main/java/io/kadai/task/internal/TaskServiceImpl.java
```diff
@@ -2589,4 +2596,160 @@ private boolean checkEditTasksPerm(TaskSummary task) {
...
+        updateTask(newTaskImpl);
+      } catch (KadaiException e) {
+        bulkLog.addError(taskId, e);
+      }
+    }
+
+    return bulkLog;
+  }
+
+  public static TaskImpl toTaskImpl(TaskPatch patch, TaskImpl task) {
```
bruderj15: unfortunate name


```suggestion
  public static TaskImpl applyTaskPatch(TaskPatch patch, TaskImpl task) {
```

rest/kadai-rest-spring/src/main/java/io/kadai/task/rest/models/TaskPatchRepresentationModel.java
```diff
@@ -0,0 +1,498 @@
...
+import io.kadai.workbasket.rest.models.WorkbasketSummaryRepresentationModel;
+import io.swagger.v3.oas.annotations.media.Schema;
+import java.time.Instant;
+import java.util.List;
+
+@Schema(
+    description =
+        "Patch model for Task with all fields nullable - null values are ignored during updates. "
+            + "Note: Fields that are not supposed to be updated manually by user e.g modified will"
+            + " be ignored.")
```
bruderj15: outdated note

rest/kadai-rest-spring/src/main/java/io/kadai/task/rest/TaskController.java
```diff
@@ -531,6 +535,30 @@ public ResponseEntity<TaskRepresentationModel> updateTask(
...
+  @Transactional(rollbackFor = Exception.class)
+  public ResponseEntity<BulkOperationResultsRepresentationModel> bulkUpdateTasks(
+      @RequestBody TaskBulkUpdateRepresentationModel requestModel) {
+
+    if (requestModel.getTaskIds() == null || requestModel.getTaskIds().isEmpty()) {
+      return ResponseEntity.ok().build();
+    }
+    if (requestModel.getFieldsToUpdate() == null) {
+      return ResponseEntity.ok().build();
+    }
```
bruderj15: Let's dig into this again, from a frontend dev view:

When you - as a frontend dev - build a UI, imagine a Task-List where the user can select tasks and then bulk-apply changes (think GitLab Issues - they allow exactly this), it may happen that a user selects **no** task in the list but still wants to apply changes.
The Payload sent contains an **empty list** of tasks.
This is a valid operation and we want to short-circuit it as a no-op with 200. Great.

If the value instead of the empty list is `null` though, this clearly indicates, that you (the frontend-dev) sent a wrong payload - the user cannot create a _null_ list - only an empty list - therefore dev-error.
This is an invalid operation and _should_ be responded to as a 400.

There's a difference between 
- _no resources_ (empty list - intentionally none)
- _no **information** about resources_ (null-value - dev-error)

This makes the goal for the task-ids clear.

---

Regarding the second if-clause:
Similar here, there's a difference between:
- applying an empty change (think empty element in a monoid)
- applying nothing (think bottom-value (cs null))

The first is valid, the latter invalid.

rest/kadai-rest-spring/src/main/java/io/kadai/task/rest/models/TaskPatchRepresentationModel.java
jamesrdi: What do you mean outdated note?

rest/kadai-rest-spring/src/main/java/io/kadai/task/rest/TaskController.java
jamesrdi: Ok, so if an empty list of taskIds or FieldstoUpdate is sent, then we should return 200, and if null we should return 400, right?

bruderj15: yes!

rest/kadai-rest-spring/src/main/java/io/kadai/task/rest/models/TaskPatchRepresentationModel.java
bruderj15: It's no longer true due to your refactoring, isn't it?

jamesrdi: It is true, fields like "modified" will be ignored since it does not exist in TaskPatchRepresentationModel and therefore will not be mapped

bruderj15: It's not true :)
Ignoring something implies it's there. 
Can't ignore your wife yelling when you don't have one (unless you're schizophrenic).

The type contains no field `modified` - hence it's impossible to ignore it.
This comment may confuse the reader (it certainly would confuse me).
Simply remove it.

jamesrdi: Okay I will remove it

rest/kadai-rest-spring/src/main/java/io/kadai/task/rest/models/TaskPatchRepresentationModel.java
```diff
@@ -0,0 +1,539 @@
...
+        && note == null
+        && description == null
+        && classificationSummary == null
+        && workbasketSummary == null
+        && businessProcessId == null
+        && parentBusinessProcessId == null
+        && primaryObjRef == null
+        && manualPriority == null
+        && isRead == null
+        && (secondaryObjectReferences == null || secondaryObjectReferences.isEmpty())
```
bruderj15: This should be given a thought:
The `.. == null || empty` check implies that one cannot clear secondary object-references and other collection-fields.

I suggest to check what the codebase is doing on other occasions then consulting the PO with that knowledge and the here popped-up question.

rest/kadai-rest-spring/src/main/java/io/kadai/task/rest/TaskController.java
```diff
@@ -531,6 +535,35 @@ public ResponseEntity<TaskRepresentationModel> updateTask(
...
+    if (requestModel.getFieldsToUpdate() == null) {
+      throw new InvalidArgumentException("fieldsToUpdate must not be null");
+    }
+    if (requestModel.getTaskIds().isEmpty()) {
+      return ResponseEntity.ok().build();
+    }
+    if (requestModel.getFieldsToUpdate().isEmpty()) {
+      return ResponseEntity.ok().build();
+    }
```
bruderj15: nitpick: merge those via disjunction

rest/kadai-rest-spring/src/main/java/io/kadai/task/rest/models/TaskPatchRepresentationModel.java
jamesrdi: You're right, I have consulted, secondaryObjectReferences and customAttributes can be emptied, but callbackInfo cannot

rest/kadai-rest-spring/src/test/java/io/kadai/task/rest/TaskControllerIntTest.java
```diff
@@ -2240,6 +2241,403 @@ void should_ThrowError_When_UpdatingTaskWithBadAttachment() {
...
+              .toEntity(TaskRepresentationModel.class);
+
+      assertThat(taskResponse.getBody()).isNotNull();
+      assertThat(taskResponse.getBody().getName()).isEqualTo("Mixed Scenario Test");
+      assertThat(taskResponse.getBody().getDescription())
+          .isEqualTo("Testing various failure scenarios");
+    }
+
+    @Test
+    void should_ThrowBadRequest_When_TaskIdsAreNull_Or_ReturnOk_When_Empty() {
```
bruderj15: please separate these cases

rest/kadai-rest-spring/src/test/java/io/kadai/task/rest/TaskControllerIntTest.java
```diff
@@ -2240,6 +2241,403 @@ void should_ThrowError_When_UpdatingTaskWithBadAttachment() {
...
+              .contentType(MediaType.APPLICATION_JSON)
+              .body(requestBody)
+              .retrieve()
+              .toEntity(BULK_RESULT_TASKS_MODEL_TYPE);
+
+      assertThat(responseEmpty.getStatusCode()).isEqualTo(HttpStatus.OK);
+    }
+
+    @Test
+    void should_ThrowBadRequest_When_FieldsToUpdateIsNull_Or_ReturnOk_When_Empty() {
```
bruderj15: more separation

bruderj15 (commented):
Last round, I guess!

bruderj15 (approved):
We got it!

PR #1346, opened 2025-09-05
Closes #512 - Test KadaiEngine with different ConnectionManagementModes

### LINKED ISSUE #512: Test KadaiEngine with different `ConnectionManagementMode`s

bruderj15:
Discovered #1347.

## Definition of Done

  - will be squashed on merge


lib/kadai-core-test/src/test/java/acceptance/common/KadaiEngineModesTest.java
```diff
@@ -0,0 +1,254 @@
...
+    void setupKadaiEngines() throws Exception {
+      String schemaName = TestContainerExtension.determineSchemaName();
+      DataSource dataSource = TestContainerExtension.DATA_SOURCE;
+      KadaiConfiguration kadaiConfiguration =
+          new KadaiConfiguration.Builder(dataSource, true, schemaName, false)
+              .initKadaiProperties()
+              .build();
+      thisKadaiEngineProxy =
+          new KadaiEngineProxy(
+              SpringKadaiEngine.buildKadaiEngine(
```
CRoberto1926, line 63: The only reason for a `SpringKadaiEngine` here is that you need a Spring managed transaction in the tests, right?

lib/kadai-core-test/src/test/java/acceptance/common/KadaiEngineTest.java
```diff
@@ -25,27 +25,61 @@
...
+    kadaiEngine.setConnectionManagementMode(mode);
+    kadaiEngine.setConnection(kadaiEngine.getConfiguration().getDataSource().getConnection());
+
+    assertThat(kadaiEngine.getConnectionManagementMode())
+        .isEqualTo(ConnectionManagementMode.EXPLICIT);
+  }
+
+  @Test
+  void should_SetModeParticipate_When_CloseConnectionForModeExplicit() {
+    kadaiEngine.closeConnection();
```
CRoberto1926, line 81: Shouldn't you set mode to `EXPLICIT` before closing the connection here?

lib/kadai-core-test/src/test/java/acceptance/common/TestUserMapper.java
```diff
@@ -0,0 +1,107 @@
...
+  @Result(property = "id", column = "USER_ID")
+  @Result(property = "firstName", column = "FIRST_NAME")
+  @Result(property = "lastName", column = "LAST_NAME")
+  @Result(property = "longName", column = "LONG_NAME")
+  TestUser findById(String id);
+
+  @Insert("INSERT INTO USER_INFO ( " + USER_INFO_COLUMNS + ") VALUES(" + USER_INFO_VALUES + ")")
+  void insert(TestUser user);
+
+  class TestUser {
```
CRoberto1926, line 50: If you prefer you can do this:

`  record TestUser(String id, String firstName, String lastName, String longName) {

    @Override
    public @NotNull String toString() {
      return "TestUser [id=" + id + ", firstName=" + firstName + ", lastName=" + lastName
          + ", longName=" + longName + "]";
    }

    @Override
    public boolean equals(Object obj) {
      if (this == obj) {
        return true;
      }
      if (obj == null) {
        return false;
      }
      if (getClass() != obj.getClass()) {
        return false;
      }
      TestUser other = (TestUser) obj;
      return Objects.equals(id, other.id) && Objects.equals(firstName, other.firstName)
          && Objects.equals(lastName, other.lastName) && Objects.equals(longName, other.longName);
    }
  }
`

lib/kadai-core-test/src/test/java/acceptance/common/KadaiEngineModesTest.java
bruderj15: Yes!

lib/kadai-core-test/src/test/java/acceptance/common/TestUserMapper.java
bruderj15: I don't really care for the test tbh

CRoberto1926 (approved):

PR #1416, opened 2025-10-06
Closes #1415: Data field from user_info table not updated correctly by the UserInfoRefreshJob

### LINKED ISSUE #1415: Data field from user_info table not updated correctly by the UserInfoRefreshJob
The data field in the user_info table is not updated correctly by the UserInfoRefreshJob. When the job runs, it overwrites the value set by the RefreshUserPostProcessor SPI with the previously stored (outdated) value.

Steps to Reproduce
1. Retrieve a user via the /users endpoint who has a non-null value in the data field.
2. Manually remove the data value from the database for this user. 
3. Run the UserInfoRefreshJob.
4. Retrieve the same user again via the /users endpoint.

Expected Behavior
The data field should be re-populated with the new value set by the RefreshUserPostProcessor SPI after the job runs.

jamesrdi:
## Definition of Done

  - will be squashed on merge


bruderj15:
Already fixed with 10.1.0

PR #1431, opened 2025-10-10
Closes #1304: Fix Release-Workflow attempting release twice

### LINKED ISSUE #1304: The new Release-Workflow leads to two release-attempts
We now manually prepare a release by bumping to the desired release-version before tagging.
This obviously triggers a CI-Pipeline due to Push to Master.

While in theory this push _should not_ result in deploying the new version to Sonatype-Central, in practice it does.
When then the CI-Pipeline from tagging tries to deploy to Sonatype-Central, an error is thrown because such a version already exists - from the previous deployment after pushing to Master for preparing the release.

This _should_ not happen because we dynamically choose the Maven-Profile depending on the `GITHUB_REF`:
https://github.com/kadai-io/kadai/blob/6296aa8404ed218813600793f334f6cdd99be3cc/.github/workflows/continuous-integration.yml#L469-L471

jamesrdi:
## Definition of Done

  - will be squashed on merge

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -472,7 +472,40 @@ jobs:
         run: ci/change_version.sh -m .
       - name: Release artifacts to Sonatype-Central
         run: |
-          ./mvnw -B deploy -P $([[ "$GITHUB_REF" =~ ^refs/tags/v[0-9]+\.[0-9]+\.[0-9]+$ ]] && echo "release" || echo "snapshot") \
+          VERSION=$(./mvnw help:evaluate -Dexpression=project.version -q -DforceStdout)
+          echo "Detected version: $VERSION"
+          
+          # Validate version is not empty/null
+          if [[ -z "$VERSION" || "$VERSION" == "null" ]]; then
+          echo "CRITICAL ERROR: Version is empty or null!"
+          echo "   Detected version: '$VERSION'"
+          echo "   Expected:"
+          echo "     - Master branch: x.x.x-SNAPSHOT"
+          echo "     - Tag (vx.x.x): x.x.x"
+          echo "   Possible causes:"
+          echo "     - Incorrect tag format"
+          echo "     - Missing or invalid <version> in pom.xml"
+          exit 1
+          fi
+          
+          if [[ "$GITHUB_REF" == "refs/heads/master" && "$VERSION" == *-SNAPSHOT ]]; then
+            echo "Deploying as SNAPSHOT"
+            PROFILE="snapshot"
+          elif [[ "$GITHUB_REF" =~ ^refs/tags/v[0-9]+\.[0-9]+\.[0-9]+$ && "$VERSION" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
+            echo "Deploying as RELEASE"
+            PROFILE="release"
+          else
+            echo "NOT ALLOWED: Invalid combination"
+            echo "   GITHUB_REF: $GITHUB_REF"
+            echo "   Version: $VERSION"
+            echo "   Expected:"
+            echo "     - Master branch + SNAPSHOT version (x.x.x-SNAPSHOT), OR"
+            echo "     - Tag (vx.x.x) + RELEASE version (x.x.x)"
+            exit 1
+          fi
+          
+          echo "Using Maven profile: $PROFILE"
+          ./mvnw -B deploy -P $PROFILE \
           --settings ci/mvnsettings.xml -DskipTests -Dcheckstyle.skip -Djacoco.skip \
           -pl :kadai-parent,\
           :kadai-common-parent,:kadai-common-logging,:kadai-common,:kadai-common-security,\


bruderj15 (commented):
We wouldn't release snapshots with the current approach anymore :(

jamesrdi:
Do you maybe have any idea how to test this, because then we need to push to master to also test if the release to sonatype got triggered or not, which will not look good

> We wouldn't release snapshots with the current approach anymore :(

Yes I just realised that, let me look into it again

.github/workflows/continuous-integration.yml
```diff
@@ -424,6 +424,85 @@ jobs:
           git commit -m "Add/update copyright headers" || echo "No changes to commit"
           git push
 
+  release_snapshots:
```
jamesrdi: Added a new job to allow release of snapshots, with a check that version is snapshot

.github/workflows/continuous-integration.yml
```diff
@@ -424,10 +424,89 @@ jobs:
...
+        run: |
+          VERSION=$(./mvnw help:evaluate -Dexpression=project.version -q -DforceStdout)
+          if [[ "$VERSION" != *-SNAPSHOT ]]; then
+            echo "ERROR: Master branch should have SNAPSHOT versions, but found: $VERSION"
+            echo "This usually happens when preparing a release. Snapshot deployment skipped."
+            exit 1
+          else
+            echo "Version check passed: $VERSION"
+            echo "VALID=true" >> $GITHUB_OUTPUT
+          fi
```
bruderj15: Instead of adding a whole another job, what about just using this conditional statement for step ` Release artifacts to Sonatype-Central`?

Instead of of selecting maven profile by tag-name like we do currently: 
```yaml
- name: Release artifacts to Sonatype-Central
        run: |
          ./mvnw -B deploy -P $([[ "$GITHUB_REF" =~ ^refs/tags/v[0-9]+\.[0-9]+\.[0-9]+$ ]] && echo "release" || echo "snapshot") \
          --settings ci/mvnsettings.xml -DskipTests -Dcheckstyle.skip -Djacoco.skip \
          -pl :kadai-parent, ...
```

We could branch on the evaluated maven property like done here and then instead of error-echoing simply deploy as release:

Pseudo:
```yaml
- name: Release artifacts to Sonatype-Central
        id: version-check
        run: |
          VERSION=$(./mvnw help:evaluate -Dexpression=project.version -q -DforceStdout)
          if [[ "$VERSION" == *-SNAPSHOT ]]; then
             ./mvnw -B deploy -P snapshot
          else
             ./mvnw -B deploy -P release
          fi
```

bruderj15 (commented):
Great idea with the evaluation of the maven-property!

.github/workflows/continuous-integration.yml
jamesrdi: Great Idea!

.github/workflows/continuous-integration.yml
```diff
@@ -472,15 +472,32 @@ jobs:
...
-          :kadai-lib-parent,:kadai-core,:kadai-spring,\
-          :kadai-rest-parent,:kadai-web,:kadai-rest-spring,\
-          :kadai-history-parent,:kadai-simplehistory-provider,:kadai-simplehistory-rest-spring,:kadai-loghistory-provider,\
-          :kadai-routing-parent,:kadai-spi-routing-dmn-router,:kadai-routing-rest
+          VERSION=$(./mvnw help:evaluate -Dexpression=project.version -q -DforceStdout)
+          echo "Detected version: $VERSION"
+          
+          if [[ "$VERSION" == *-SNAPSHOT ]]; then
+            echo "Deploying as SNAPSHOT"
+            ./mvnw -B deploy -P snapshot \
```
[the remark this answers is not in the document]
bruderj15: @jamesrdi yeah I'd have also remarked that. Please fail if empty/null.

.github/workflows/continuous-integration.yml
```diff
@@ -472,15 +472,32 @@ jobs:
...
-          :kadai-common-parent,:kadai-common-logging,:kadai-common,:kadai-common-security,\
-          :kadai-common-data,:kadai-common-test,:kadai-test-api,\
-          :kadai-lib-parent,:kadai-core,:kadai-spring,\
-          :kadai-rest-parent,:kadai-web,:kadai-rest-spring,\
-          :kadai-history-parent,:kadai-simplehistory-provider,:kadai-simplehistory-rest-spring,:kadai-loghistory-provider,\
-          :kadai-routing-parent,:kadai-spi-routing-dmn-router,:kadai-routing-rest
+          VERSION=$(./mvnw help:evaluate -Dexpression=project.version -q -DforceStdout)
+          echo "Detected version: $VERSION"
+          
+          if [[ "$VERSION" == *-SNAPSHOT ]]; then
```
jamesrdi: I think here we should not only check for the version name if it contains snapshot, but we also have to check the github ref, if it is refs/heads/master or refs/tags/v, because with the current approach, when bumping and then pushing to master, the artifact will still be released, and later the error is thrown when tagging

.github/workflows/continuous-integration.yml
```diff
@@ -472,15 +472,46 @@ jobs:
...
+            --settings ci/mvnsettings.xml -DskipTests -Dcheckstyle.skip -Djacoco.skip \
+            -pl :kadai-parent,\
+            :kadai-common-parent,:kadai-common-logging,:kadai-common,:kadai-common-security,\
+            :kadai-common-data,:kadai-common-test,:kadai-test-api,\
+            :kadai-lib-parent,:kadai-core,:kadai-spring,\
+            :kadai-rest-parent,:kadai-web,:kadai-rest-spring,\
+            :kadai-history-parent,:kadai-simplehistory-provider,:kadai-simplehistory-rest-spring,:kadai-loghistory-provider,\
+            :kadai-routing-parent,:kadai-spi-routing-dmn-router,:kadai-routing-rest
+          else
+            echo " NOT ALLOWED: Invalid combination"
```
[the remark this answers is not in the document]
bruderj15: @jamesrdi fix the formatting and we're gonna give it a go, looks good! :)

.github/workflows/continuous-integration.yml
```diff
@@ -472,15 +472,46 @@ jobs:
...
+            echo "Deploying as RELEASE"
+            ./mvnw -B deploy -P release \
+            --settings ci/mvnsettings.xml -DskipTests -Dcheckstyle.skip -Djacoco.skip \
+            -pl :kadai-parent,\
+            :kadai-common-parent,:kadai-common-logging,:kadai-common,:kadai-common-security,\
+            :kadai-common-data,:kadai-common-test,:kadai-test-api,\
+            :kadai-lib-parent,:kadai-core,:kadai-spring,\
+            :kadai-rest-parent,:kadai-web,:kadai-rest-spring,\
+            :kadai-history-parent,:kadai-simplehistory-provider,:kadai-simplehistory-rest-spring,:kadai-loghistory-provider,\
+            :kadai-routing-parent,:kadai-spi-routing-dmn-router,:kadai-routing-rest
```
bruderj15: One more improvement:
Can we simply store the selected profile in a var and then pass the var as arg to the deploy-fun?
This way we dont have to replicate the listed modules.

bruderj15 (approved):
Let's get it rolling!

PR #1458, opened 2025-10-22
Closes #1433: Fix Inefficient deployment of artifacts to Sonatype Central in CI pipeline

### LINKED ISSUE #1433: Inefficient deployment of artifacts to Sonatype Central in CI pipeline
Maven deploy is taking multiple minutes despite artifacts being pre-built and downloaded from a previous job. The deployment should only upload existing artifacts to Sonatype Central, but Maven is re-running unnecessary build lifecycle phases (compilation, packaging) and plugin executions.

https://github.com/kadai-io/kadai/blob/e11cc6a2adfbba84a1cd7812247a51de6d1bb16b/.github/workflows/continuous-integration.yml#L475-L483

## Potential Fixes

*Claude recommended to me the following potential fixes (have to be checked and tested):*

**Artifact path conflicts:**

1. **Fix cache/artifact overlap** - Downloading artifacts to `~/.m2/repository/io/kadai` while caching `~/.m2` causes conflicts. Either exclude kadai artifacts from cache or use module `target/` directories instead
2. **Use `**/target/*.jar` pattern** - Upload/download artifacts from module target directories where Maven naturally expects them, rather than to `.m2/repository`

**Maven optimization (try after fixing paths):**

3. **Use `deploy:deploy` goal directly** - Skips entire build lifecycle: `./mvnw -B deploy:deploy` instead of `./mvnw -B deploy`
4. **Add skip flags** - Prevent compilation/packaging: `-Dmaven.main.skip=true -Dmaven.compiler.skip=true -Dmaven.resources.skip=true -Dmaven.install.skip=true`
5. **Parallel deployment** - Add `-T 4` for multi-threaded deployment (if modules are independent)

jamesrdi:
## Definition of Done

  - will be squashed on merge

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -519,7 +519,7 @@ jobs:
           fi
           
           echo "Using Maven profile: $PROFILE"
-          ./mvnw -B deploy -P $PROFILE \
+          ./mvnw -B deploy:deploy -P $PROFILE \
           --settings ci/mvnsettings.xml \
           -pl :kadai-parent,\
           :kadai-common-parent,:kadai-common-logging,:kadai-common,:kadai-common-security,\
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -48,7 +48,9 @@ jobs:
         id: cache
         uses: actions/cache@v4
         with:
-          path: ~/.m2
+          path: |
+            ~/.m2
+            !~/.m2/repository/io/kadai
           key: ${{ runner.os }}-${{ env.CACHE_MAVEN_NAME }}-${{ hashFiles('**/pom.xml') }}
           restore-keys: ${{ runner.os }}-${{ env.CACHE_MAVEN_NAME }}
       - name: Change versions to match tag
@@ -99,7 +101,9 @@ jobs:
         id: maven-cache
         uses: actions/cache@v4
         with:
-          path: ~/.m2
+          path: |
+            ~/.m2
+            !~/.m2/repository/io/kadai
           key: ${{ runner.OS }}-${{ env.CACHE_MAVEN_FOR_WEB_NAME }}-${{ hashFiles('**/pom.xml') }}
           restore-keys: ${{ runner.OS }}-${{ env.CACHE_MAVEN_FOR_WEB_NAME }}
       - name: Populate maven cache
@@ -157,7 +161,9 @@ jobs:
         id: maven-cache
         uses: actions/cache@v4
         with:
-          path: ~/.m2
+          path: |
+            ~/.m2
+            !~/.m2/repository/io/kadai
           key: ${{ runner.OS }}-${{ env.CACHE_MAVEN_FOR_WEB_NAME }}-${{ hashFiles('**/pom.xml') }}
           restore-keys: ${{ runner.OS }}-${{ env.CACHE_MAVEN_FOR_WEB_NAME }}
       # Theoretically this step below not necessary because we reuse the cache from the 'compile_frontend' job.
@@ -206,7 +212,9 @@ jobs:
         id: maven-cache
         uses: actions/cache@v4
         with:
-          path: ~/.m2
+          path: |
+            ~/.m2
+            !~/.m2/repository/io/kadai
           key: ${{ runner.os }}-${{ env.CACHE_MAVEN_NAME }}-${{ hashFiles('**/pom.xml') }}
       - name: Download kadai artifacts
         uses: actions/download-artifact@v5
@@ -303,7 +311,9 @@ jobs:
       - name: Cache maven dependencies
         uses: actions/cache@v4
         with:
-          path: ~/.m2
+          path: |
+            ~/.m2
+            !~/.m2/repository/io/kadai
           key: ${{ runner.os }}-${{ env.CACHE_MAVEN_NAME }}-${{ hashFiles('**/pom.xml') }}
           restore-keys: ${{ runner.os }}-${{ env.CACHE_MAVEN_NAME }}
       - name: Download kadai artifacts
@@ -353,7 +363,9 @@ jobs:
       - name: Cache maven dependencies
         uses: actions/cache@v4
         with:
-          path: ~/.m2
+          path: |
+            ~/.m2
+            !~/.m2/repository/io/kadai
           key: ${{ runner.os }}-${{ env.CACHE_MAVEN_NAME }}-${{ hashFiles('**/pom.xml') }}
           restore-keys: ${{ runner.os }}-${{ env.CACHE_MAVEN_NAME }}
       - name: Download kadai artifacts
@@ -451,7 +463,9 @@ jobs:
       - name: Cache maven dependencies
         uses: actions/cache@v4
         with:
-          path: ~/.m2
+          path: |
+            ~/.m2
+            !~/.m2/repository/io/kadai
           key: ${{ runner.os }}-${{ env.CACHE_MAVEN_NAME }}-${{ hashFiles('**/pom.xml') }}
           restore-keys: ${{ runner.os }}-${{ env.CACHE_MAVEN_NAME }}
       - name: Download kadai artifacts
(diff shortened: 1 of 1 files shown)


.github/workflows/continuous-integration.yml
```diff
@@ -507,13 +503,17 @@ jobs:
...
           ./mvnw -B deploy -P $PROFILE \
           --settings ci/mvnsettings.xml -DskipTests -Dcheckstyle.skip -Djacoco.skip \
+          -Dmaven.main.skip=true -Dmaven.compiler.skip=true -Dmaven.resources.skip=true \
+          -Dmaven.install.skip=true \
           -pl :kadai-parent,\
           :kadai-common-parent,:kadai-common-logging,:kadai-common,:kadai-common-security,\
           :kadai-common-data,:kadai-common-test,:kadai-test-api,\
           :kadai-lib-parent,:kadai-core,:kadai-spring,\
           :kadai-rest-parent,:kadai-web,:kadai-rest-spring,\
+          :kadai-rest-spring-test-lib,:kadai-rest-spring-example-common,:kadai-rest-spring-example-boot,\
```
bruderj15: Why did you add these? These should not be deployed.

.github/workflows/continuous-integration.yml
```diff
@@ -507,13 +503,17 @@ jobs:
           echo "Using Maven profile: $PROFILE"
           ./mvnw -B deploy -P $PROFILE \
           --settings ci/mvnsettings.xml -DskipTests -Dcheckstyle.skip -Djacoco.skip \
```
davidkopp: `-Dcheckstyle.skip` and `-Djacoco.skip` are probably unnecessary when skipping compilation

.github/workflows/continuous-integration.yml
```diff
@@ -507,13 +503,17 @@ jobs:
...
+          -Dmaven.main.skip=true -Dmaven.compiler.skip=true -Dmaven.resources.skip=true \
```
davidkopp: `-Dmaven.main.skip=true` is a property of the maven-compiler-plugin. Because we use `-Dmaven.compiler.skip=true` I assume this is not necessary and can be removed.

.github/workflows/continuous-integration.yml
```diff
@@ -66,8 +66,6 @@ jobs:
           name: ${{ env.ARTIFACTS_KADAI_JARS_NAME }}
           path: ${{ env.ARTIFACTS_KADAI_JARS_PATH }}
           if-no-files-found: error
-      - name: Remove kadai artifacts from cache
```
davidkopp, line 69: Why did you remove this? I think the removal of the from the cached `.m2` makes sense as we don't want to pollute the cache with KADAI build artifacts.

.github/workflows/continuous-integration.yml
```diff
@@ -122,8 +120,6 @@ jobs:
           name: ${{ env.ARTIFACTS_KADAI_WEB_NAME }}
           path: ${{ env.ARTIFACTS_KADAI_WEB_PATH }}
           if-no-files-found: error
-      - name: Remove kadai artifacts from cache
-        run: rm -rf ~/.m2/repository/io/kadai
```
davidkopp, line 126: Same here. I think the removal makes sense. But you could replace the hardcoded path with `${{ env.ARTIFACTS_KADAI_JARS_PATH }}` to be consistent with the `compile_backend` job.

.github/workflows/continuous-integration.yml
davidkopp: Have you tested `./mvnw -B deploy:deploy`? If I understand it correctly, it skips the entire build lifecycle already, so not additional skip flags would be necessary.

davidkopp (changes requested):

.github/workflows/continuous-integration.yml
jamesrdi: No I haven't, but I will try it out

.github/workflows/continuous-integration.yml
```diff
@@ -507,13 +503,17 @@ jobs:
...
+          -Dmaven.install.skip=true \
           -pl :kadai-parent,\
           :kadai-common-parent,:kadai-common-logging,:kadai-common,:kadai-common-security,\
           :kadai-common-data,:kadai-common-test,:kadai-test-api,\
           :kadai-lib-parent,:kadai-core,:kadai-spring,\
           :kadai-rest-parent,:kadai-web,:kadai-rest-spring,\
+          :kadai-rest-spring-test-lib,:kadai-rest-spring-example-common,:kadai-rest-spring-example-boot,\
           :kadai-history-parent,:kadai-simplehistory-provider,:kadai-simplehistory-rest-spring,:kadai-loghistory-provider,\
-          :kadai-routing-parent,:kadai-spi-routing-dmn-router,:kadai-routing-rest
+          :kadai-routing-parent,:kadai-spi-routing-dmn-router,:kadai-routing-rest \
```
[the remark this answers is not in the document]
jamesrdi: I do not understand this

davidkopp:
@jamesrdi In the last commit you removed all skips from the maven deploy command. Why? Now the whole build and test is run again.

Here a version with some skips removed but not all (**untested**):
```
./mvnw -B deploy -P $PROFILE \
  --settings ci/mvnsettings.xml \
  -DskipTests \
  -Dmaven.compiler.skip=true \
  -Dmaven.resources.skip=true \
  -Dmaven.install.skip=true \
  ...
```

Or you may try `deploy:deploy` (**untested**):
```
./mvnw -B deploy:deploy -P $PROFILE \
  --settings ci/mvnsettings.xml \
  ...
```

jamesrdi:
> @jamesrdi In the last commit you removed all skips from the maven deploy command. Why? Now the whole build and test is run again.
> 
> Here a version with some skips removed but not all (**untested**):
> 
> ```
> ./mvnw -B deploy -P $PROFILE \
>   --settings ci/mvnsettings.xml \
>   -DskipTests \
>   -Dmaven.compiler.skip=true \
>   -Dmaven.resources.skip=true \
>   -Dmaven.install.skip=true \
>   ...
> ```
> 
> Or you may try `deploy:deploy` (**untested**):
> 
> ```
> ./mvnw -B deploy:deploy -P $PROFILE \
>   --settings ci/mvnsettings.xml \
>   ...
> ```

Sorry I wanted to try deploy:deploy but forgot to change it

.github/workflows/continuous-integration.yml
davidkopp: @jamesrdi If you use `!~/.m2/repository/io/kadai` now (what I like!) you can remove this step now.

.github/workflows/continuous-integration.yml
davidkopp, line 438: remember that this condition is false, so this job does not get triggered. Either you test your changes locally or you need to change the condition here in the pipeline and clean up after (e.g. deleting a snapshot from sonatype, don't know if that is feasible for you)

jamesrdi: We are going to trigger a release to sonatype to test this job

.github/workflows/continuous-integration.yml
```diff
@@ -505,15 +515,16 @@ jobs:
...
+          --settings ci/mvnsettings.xml \
           -pl :kadai-parent,\
           :kadai-common-parent,:kadai-common-logging,:kadai-common,:kadai-common-security,\
           :kadai-common-data,:kadai-common-test,:kadai-test-api,\
           :kadai-lib-parent,:kadai-core,:kadai-spring,\
           :kadai-rest-parent,:kadai-web,:kadai-rest-spring,\
           :kadai-history-parent,:kadai-simplehistory-provider,:kadai-simplehistory-rest-spring,:kadai-loghistory-provider,\
-          :kadai-routing-parent,:kadai-spi-routing-dmn-router,:kadai-routing-rest
+          :kadai-routing-parent,:kadai-spi-routing-dmn-router,:kadai-routing-rest \
+          -T 4
```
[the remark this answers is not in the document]
jamesrdi: Do we want to try multi threads or not?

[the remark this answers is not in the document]
bruderj15: I don't think we should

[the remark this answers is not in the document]
davidkopp: I agree, for the deploy command it makes no sense. However, multi-threading could make sense for the build and test jobs.
fyi: Mohamed is currently working on a PR that improves the build and test execution in the pipeline.

davidkopp (changes requested):
As Copilot mentioned, `-T4` makes no sense for the deployment. That was a problematic proposal from Claude. That should be removed.
The rest looks good for me.

davidkopp (approved):

bruderj15 (approved):
Let's see whether this works!

bruderj15:
@jamesrdi failed!
See: https://github.com/kadai-io/kadai/actions/runs/19501141274/job/55816012784#step:10:253
Please get this fixed asap.

PR #1520, opened 2025-11-20
Closes #1511 - Improve build & tests in pipeline

MohamedSellami9:
## Definition of Done

- [x] The corresponding ticket is in state `In Review`
- [x] The corresponding ticket is attached to this Pull-Request
- [x] The commit-message-format `Closes #ISSUE_ID - PROBLEM/SOLUTION` 
  - is present as single-commit already or
  - will be squashed on merge
- [ ] The changes pass the SonarQubeCloud-Quality-Gate
- [x] If the documentation needs an update, following there's a link to it:
- [x] If extra release-notes are required, they're listed indented below:

PR #1533, opened 2025-11-27
Partially closes #1511 - Parallel build and e2e conditional execution

### LINKED ISSUE #1511: Inefficient build & test execution in CI pipeline
Whenever there is a change of a file, all build and test jobs are executed. That is highly inefficient. The jobs should only be triggered if a relevant file was changed (e.g. if there was only a change of a markdown file the pipeline should not be executed). Also if there is only a change in a sub-module (e.g. the spring example application), the test execution of other modules is unnecessary.

@MohamedSellami9 is currently working on a PR for this.

MohamedSellami9:
## Definition of Done

  - will be squashed on merge
- [ ] If extra release-notes are required, they're listed indented below:


MohamedSellami9:
This Pull request is part of #1520, which introduces parallel builds and conditional E2E execution. This portion can be discussed and merged separately, as the conditional testing changes in #1520 may require more time to be discussed.

.github/workflows/continuous-integration.yml
```diff
@@ -243,7 +279,13 @@ jobs:
   test_e2e:
     runs-on: ubuntu-22.04
     name: Test E2E
-    needs: [ compile_frontend, compile_backend ]
+    needs: [detect_changes, test_frontend, test_backend_17, test_backend_21]
```
bruderj15: This is redundant due to transitive closure, right?

```suggestion
    needs: [test_frontend, test_backend_17, test_backend_21]
```

.github/workflows/continuous-integration.yml
```diff
@@ -243,7 +279,13 @@ jobs:
...
+    if: |
+      always() &&
```
bruderj15, line 286: the always-thingy is required due to github-woowoo?

.github/workflows/continuous-integration.yml
```diff
@@ -95,6 +129,8 @@ jobs:
   compile_frontend:
     name: Compile kadai-web
     runs-on: ubuntu-22.04
+    needs: [ detect_changes ]
+    if: needs.detect_changes.outputs.backend_changed == 'true' || needs.detect_changes.outputs.frontend_changed == 'true'
```
bruderj15: Why check for backend changed here?

```suggestion
    if: needs.detect_changes.outputs.frontend_changed == 'true'
```

.github/workflows/continuous-integration.yml
```diff
@@ -243,7 +279,13 @@ jobs:
...
     runs-on: ubuntu-22.04
     name: Test E2E
-    needs: [ compile_frontend, compile_backend ]
+    needs: [detect_changes, test_frontend, test_backend_17, test_backend_21]
+    if: |
+      always() &&
+      (needs.detect_changes.outputs.frontend_changed == 'true' || needs.detect_changes.outputs.backend_changed == 'true') &&
+      (needs.test_frontend.result == 'success' || needs.test_frontend.result == 'skipped') &&
+      (needs.test_backend_17.result == 'success' || needs.test_backend_17.result == 'skipped') &&
+      (needs.test_backend_21.result == 'success' || needs.test_backend_21.result == 'skipped')
```
bruderj15: can be simplified by negating failure (I guess)

bruderj15 (commented):
In separate commits, with enough time between to run entire CI, can you temporarily trigger changes in:
- backend
- frontend
- backend and frontend

...such that we can verify deisred behavior?

.github/workflows/continuous-integration.yml
MohamedSellami9: We need to compile both frontend and backend regardless of which one changed. Otherwise, jobs:  release_artifacts , deploy_to_azure , and upload_to_sonar will fail because they always expect both frontend and backend builds to be available.

.github/workflows/continuous-integration.yml
MohamedSellami9: without the always(), if one of the dependencies in needs array is skipped (for exemple there is no changes in backend so test_backend_17 and test_backend_21 are skipped ) e2e job wil be skipped

.github/workflows/continuous-integration.yml
MohamedSellami9: yes change is made

.github/workflows/continuous-integration.yml
MohamedSellami9: This is needed because the job’s condition depends on the detect_changes output. Without using it here, the workflow would fail with an error.

bruderj15: Yeah but as the other elements of `needs: [..]` all already depend on `detect_changes` it's redundant here - hence why we can make use of the transitive closure.

We already use the closure for `test_frontend` and `test_backend_*`, don't we?
These don't directly depend on `detect_changes` but their other _needs_ - the compile-jobs - do.

.github/workflows/continuous-integration.yml
bruderj15: Certified github-woowoo. Everyone would expect result `skipped` to be the `mempty` in the monoid of job-results.

.github/workflows/continuous-integration.yml
davidkopp: I don't have a strong opinion about this. But at least we should be consistent. In `release_artifacts`, `deploy_to_azure` and `upload_to_sonar` the `detect_changes` is not part of the `needs` definition.

.github/workflows/continuous-integration.yml
```diff
@@ -33,9 +33,44 @@ env:
...
+      backend_changed: ${{ steps.filter.outputs.backend }}
+      frontend_changed: ${{ steps.filter.outputs.frontend }} 
+      code_changed: ${{ steps.filter.outputs.backend == 'true' || steps.filter.outputs.frontend == 'true' }}  
+    steps:
+      - uses: actions/checkout@v5
+      - uses: dorny/paths-filter@v3
+        id: filter
+        with:
+          base: ${{ github.ref_name }}
+          filters: |
```
davidkopp, line 50: consider adding the exclude markdown files pattern to both, backend and frontend:
`- '!**/*.md'`

davidkopp (commented):
Looks really good for me. Only two comments from my side.

.github/workflows/continuous-integration.yml
MohamedSellami9: the problem is that we are using the output of `detect_changes` in `Test E2E` so we need it in the `needs` if not we cannot access `needs.detect_changes.outputs.code_changed` or any other output

bruderj15: Fair point!

MohamedSellami9: Now i see what you mean in `deploy_to_azure` , `deploy_to_azure` and `upload_to_sonar` we are using output of `detect_changes` and it is not in needs i will verify if it is really optional
my answers was folowing the documentation [needs-context](https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#needs-context) they say "Note that this doesn't include implicitly dependent jobs"

bruderj15 (commented):
Upload to SonarQube didn't happen...?

EDIT: Ah new commits just popped up, I was looking at an old run.

.github/workflows/continuous-integration.yml
MohamedSellami9: I tried it but it did not work,
!\*\*/\*.md won't work with our current setup. With the default predicate-quantifier: 'some', exclusion patterns are ignored when files already match inclusion patterns like common/\*\* or web/\*\* (see paths-filter docs).

Using predicate-quantifier: 'every' would require all patterns to match, which would break detection entirely since a file in common/ wouldn't match lib/\*\*, history/\*\*, etc. for backend.

If the backend files are in only one subfolder, it would work using predicate-quantifier: 'every'.

MohamedSellami9:
The ci pipelines runs for testing:
[testing backend](https://github.com/MohamedSellami9/kadai/actions/runs/19751883066)
[testing frontend](https://github.com/MohamedSellami9/kadai/actions/runs/19752048891)
[testing both frontend and backend](https://github.com/MohamedSellami9/kadai/actions/runs/19752161458)

.github/workflows/continuous-integration.yml
davidkopp: Understood. As there are only a few Markdown files in the specified folders, that's not a big issue.

davidkopp (approved):

bruderj15 (approved):

PR #1536, opened 2025-11-28
Closes #1532 - Fix querying Task-Summaries with additional user-information fails on Postgres and DB2

### LINKED ISSUE #1532: Querying Task-Summaries with additional user-information fails on Postgres and DB2
## Description

When querying task-summaries via the underlying SQL in `TaskQuerySqlProvider::queryTaskSummaries` on Postgres, the query fails because `LEFT JOIN ... FOR UPDATE` is forbidden in Postgres.

The `LEFT JOIN ...` is conditionally enabled by `TaskQueryImpl::joinWithUserInfo`.
The `FETCH FIRST .. ROWS ONLY FOR UPDATE SKIP LOCKED` is conditionally enabled by `TaskQueryImpl::lockResults`.
The `FOR UPDATE SKIP LOCKED` is conditionally enabled by `TaskQueryImpl::selectAndClaim`.

## Current behavior

Query fails, MyBatis errors out.

```bash
Caught exception while trying to cancel claim referenced tasks","thrown":{"commonElementCount":0,"localizedMessage":"\n### Error querying database.  
Cause: org.postgresql.util.PSQLException: FEHLER: FOR UPDATE kann nicht auf die nullbare Seite eines Ã¤uÃŸeren Verbundes angewendet werden\n### 
The error may exist in io/kadai/task/internal/TaskQueryMapper.java (best guess)\n### 
The error may involve io.kadai.task.internal.TaskQueryMapper.queryTaskSummaries-Inline\n### The error occurred while setting parameters\n### SQL: 

SELECT   t.id, t.external_id, t.created, t.claimed, t.completed, t.modified, t.planned, t.received, t.due, t.name, t.creator, 
    t.description, t.note, t.priority, t.manual_priority, t.state, t.classification_category, t.classification_key, 
    t.classification_id, t.workbasket_id, t.workbasket_key, t.domain, t.business_process_id, t.parent_business_process_id, 
    t.owner, t.por_company, t.por_system, t.por_instance, t.por_type, t.por_value, t.is_read, t.is_transferred, t.custom_1, 
    t.custom_2, t.custom_3, t.custom_4, t.custom_5, t.custom_6, t.custom_7, t.custom_8, t.custom_9, t.custom_10, 
    t.custom_11, t.custom_12, t.custom_13, t.custom_14, t.custom_15, t.custom_16, t.custom_int_1, t.custom_int_2, 
    t.custom_int_3, t.custom_int_4, t.custom_int_5, t.custom_int_6, t.custom_int_7, t.custom_int_8, t.number_of_comments , u.LONG_NAME    
FROM TASK t  LEFT JOIN USER_INFO u ON t.owner = u.USER_ID  
WHERE  (t.CALLBACK_STATE IN(?)) AND (t.STATE IN(?))
FETCH FIRST 64 ROWS ONLY FOR UPDATE SKIP LOCKED
```

## Expected behavior

Query succeeds.

bruderj15:
## Definition of Done

  - will be squashed on merge
  - Fixed a bug where the combination of `lockResults` and `joinWithUserInfo` in `TaskQueryImpl` would alllow construction of an invalid query - leading to a runtime exception being thrown by MyBatis due to the DBMS complaining. We now forbid the combination by throwing an `IllegalArgumentException` when constructing `TaskQueryImpl`.

### DIFF
--- a/lib/kadai-core/src/main/java/io/kadai/task/internal/TaskQuerySqlProvider.java
+++ b/lib/kadai-core/src/main/java/io/kadai/task/internal/TaskQuerySqlProvider.java
@@ -55,7 +55,7 @@ public class TaskQuerySqlProvider {
         + "<if test=\"addAttachmentClassificationNameToSelectClauseForOrdering\">, "
         + "ac.NAME as ACNAME </if>"
         + "<if test=\"addWorkbasketNameToSelectClauseForOrdering\">, w.NAME as WNAME </if>"
-        + "<if test=\"joinWithUserInfo\">, u.LONG_NAME</if>"
+        + "<if test=\"joinWithUserInfo and lockResults == 0\">, u.LONG_NAME</if>"
         + groupByPorIfActive()
         + groupBySorIfActive()
         + "FROM TASK t "
@@ -74,7 +74,7 @@ public class TaskQuerySqlProvider {
         + "<if test=\"joinWithWorkbaskets\">"
         + "LEFT JOIN WORKBASKET w ON t.WORKBASKET_ID = w.ID "
         + "</if>"
-        + "<if test=\"joinWithUserInfo\">"
+        + "<if test=\"joinWithUserInfo and lockResults == 0\">"
         + "LEFT JOIN USER_INFO u ON t.owner = u.USER_ID "
         + "</if>"
         + OPENING_WHERE_TAG
@@ -327,7 +327,6 @@ public class TaskQuerySqlProvider {
         .filter(column -> column.startsWith("t"))
         .collect(Collectors.joining(", "));
   }
-
 
   private static String db2selectFields() {
     // needs to be the same order as the commonSelectFields (TaskQueryColumnValue)
--- a/lib/kadai-core/src/main/java/io/kadai/task/internal/TaskQuerySqlProvider.java
+++ b/lib/kadai-core/src/main/java/io/kadai/task/internal/TaskQuerySqlProvider.java
@@ -55,7 +55,7 @@ public class TaskQuerySqlProvider {
         + "<if test=\"addAttachmentClassificationNameToSelectClauseForOrdering\">, "
         + "ac.NAME as ACNAME </if>"
         + "<if test=\"addWorkbasketNameToSelectClauseForOrdering\">, w.NAME as WNAME </if>"
-        + "<if test=\"joinWithUserInfo and lockResults == 0\">, u.LONG_NAME</if>"
+        + "<if test=\"joinWithUserInfo\">, u.LONG_NAME</if>"
         + groupByPorIfActive()
         + groupBySorIfActive()
         + "FROM TASK t "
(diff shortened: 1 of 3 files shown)


lib/kadai-core/src/main/java/io/kadai/task/internal/TaskQuerySqlProvider.java
```diff
@@ -55,7 +55,7 @@ public static String queryTaskSummaries() {
         + "<if test=\"addAttachmentClassificationNameToSelectClauseForOrdering\">, "
         + "ac.NAME as ACNAME </if>"
         + "<if test=\"addWorkbasketNameToSelectClauseForOrdering\">, w.NAME as WNAME </if>"
-        + "<if test=\"joinWithUserInfo\">, u.LONG_NAME</if>"
+        + "<if test=\"joinWithUserInfo and lockResults == 0\">, u.LONG_NAME</if>"
```
bruderj15: This is silently disallowing the combination.
We could instead throw an exception in `TaskQueryImpl` at query-generation (pre execution).

jamesrdi:
Actually, we do not need to lock both tables USER_INFO and TASK, right? Theoretically we only need to lock the table TASK, which is not nullable. So we can use `FOR UPDATE of t` to only lock the TASK table. That way we can use `selectAndClaim` and `lockResults` in combination with `addAdditionalUserInfo`, don't you think?

jamesrdi:
> Actually, we do not need to lock both tables USER_INFO and TASK, right? Theoretically we only need to lock the table TASK, which is not nullable. So we can use `FOR UPDATE of t` to only lock the TASK table. That way we can use `selectAndClaim` and `lockResults` in combination with `addAdditionalUserInfo`, don't you think?

I just tested this for lockResultsEqual and addAdditionalUserInfo, and it worked, the error was not thrown, I believe it should also work for selectAndClaim and addAdditionalUserInfo.

bruderj15:
@jamesrdi 
> I just tested this for lockResultsEqual and addAdditionalUserInfo, and it worked, the error was not thrown, I believe it should also work for selectAndClaim and addAdditionalUserInfo.

Would you provide a new PR targetting this PR's branch with your proposed changes?

jamesrdi:
> @jamesrdi
> 
> > I just tested this for lockResultsEqual and addAdditionalUserInfo, and it worked, the error was not thrown, I believe it should also work for selectAndClaim and addAdditionalUserInfo.
> 
> Would you provide a new PR targetting this PR's branch with your proposed changes?

Yes, I can do that.

jamesrdi:
> Actually, we do not need to lock both tables USER_INFO and TASK, right? Theoretically we only need to lock the table TASK, which is not nullable. So we can use `FOR UPDATE of t` to only lock the TASK table. That way we can use `selectAndClaim` and `lockResults` in combination with `addAdditionalUserInfo`, don't you think?

Nevermind this, I just tried it out with DB2 but it did not work

jamesrdi (approved):
This PR looks good to me

PR #1606, opened 2025-12-22
Closes #1596 - Extend column-size for custom fields on task table

### LINKED ISSUE #1596: Extend size for column `Task::custom...` to allow strings of length 1023

bruderj15:
## Definition of Done

  - will be squashed on merge


common/kadai-common/src/main/resources/sql/postgres/kadai_11.0.0_to_kadai_12.0.0_schema_update_postgres.sql
```diff
@@ -1,4 +1,4 @@
-SET SCHEMA %schemaName%;
+SET search_path = %schemaName%;
```
CRoberto1926, line 1: Why this change?

bruderj15: because it previously wasn't valid postgres-syntax. Now it is.

See [example](https://www.crunchydata.com/blog/demystifying-schemas-search_path-through-examples#the-search-path) (didn't find specs)

CRoberto1926 (approved):

PR #1772, opened 2026-03-12
Closes #1769 - Add REST-Endpoint `/api/v1/bulkterminate/force`

### LINKED ISSUE #1769: Add REST-Endpoint `/api/v1/bulkterminate/force`

bruderj15:
## Definition of Done

  - will be squashed on merge

### DIFF
--- a/lib/kadai-core/src/main/java/io/kadai/task/internal/TaskMapper.java
+++ b/lib/kadai-core/src/main/java/io/kadai/task/internal/TaskMapper.java
@@ -197,6 +197,14 @@ public interface TaskMapper {
 
   @Update(
       "<script>"
+          + " UPDATE TASK SET COMPLETED = #{referenceTask.completed}, MODIFIED = #{referenceTask.modified}, STATE = #{referenceTask.state}"
+          + " WHERE ID IN <foreach item='taskId' index='index' separator=',' open='(' close=')' collection='taskIds'>#{taskId}</foreach>"
+          + "</script>")
+  void updateTerminated(
+      @Param("taskIds") List<String> taskIds, @Param("referenceTask") TaskSummary referenceTask);
+
+  @Update(
+      "<script>"
           + " UPDATE TASK SET CLAIMED = #{referenceTask.claimed}, MODIFIED = #{referenceTask.modified}, STATE = #{referenceTask.state}, OWNER = #{referenceTask.owner}, IS_READ = #{referenceTask.isRead}"
           + " WHERE ID IN <foreach item='taskId' index='index' separator=',' open='(' close=')' collection='taskIds'>#{taskId}</foreach>"
           + "</script>")
--- a/lib/kadai-core/src/main/java/io/kadai/task/internal/TaskServiceImpl.java
+++ b/lib/kadai-core/src/main/java/io/kadai/task/internal/TaskServiceImpl.java
@@ -970,6 +970,12 @@ public class TaskServiceImpl implements TaskService {
   }
 
   @Override
+  public BulkOperationResults<String, KadaiException> forceTerminateTasks(List<String> taskIds)
+      throws InvalidArgumentException, NotAuthorizedException {
+    return terminateTasks(taskIds);
+  }
+
+  @Override
   @Deprecated(forRemoval = true, since = "11.0.0")
   @SuppressWarnings({"deprecation", "removal"})
   public List<String> updateTasks(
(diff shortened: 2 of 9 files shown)


lib/kadai-core/src/main/java/io/kadai/task/internal/TaskServiceImpl.java
```diff
@@ -2412,6 +2418,74 @@ private void updateTasksToBeCompleted(Stream<TaskSummaryImpl> taskSummaries, Ins
     }
   }
 
+  private BulkOperationResults<String, KadaiException> terminateTasks(List<String> taskIds)
+      throws InvalidArgumentException, NotAuthorizedException {
+    kadaiEngine.getEngine().checkRoleMembership(KadaiRole.ADMIN, KadaiRole.TASK_ADMIN);
+    try {
+      kadaiEngine.openConnection();
+      if (taskIds == null) {
```
CRoberto1926: Personally I would also check for `taskIds.isEmpty()`, and if true, immediately return bulkLog. This has the advantage that in the rest of the method is guaranteed that there is at least one taskId to process.

lib/kadai-core/src/main/java/io/kadai/task/internal/TaskServiceImpl.java
```diff
@@ -969,6 +969,12 @@ public BulkOperationResults<String, KadaiException> forceCompleteTasks(List<Stri
     return completeTasks(taskIds, true);
   }
 
+  @Override
+  public BulkOperationResults<String, KadaiException> forceTerminateTasks(List<String> taskIds)
+      throws InvalidArgumentException, NotAuthorizedException {
+    return terminateTasks(taskIds);
```
CRoberto1926: Why did you write another method here? What was the problem with inlining `terminateTasks` here?

lib/kadai-core/src/main/java/io/kadai/task/internal/TaskServiceImpl.java
```diff
@@ -2412,6 +2418,74 @@ private void updateTasksToBeCompleted(Stream<TaskSummaryImpl> taskSummaries, Ins
...
+                  summary.setCompleted(now);
+                  summary.setState(TERMINATED);
+                  return (TaskSummary) summary;
+                })
+            .toList();
+
+    List<String> taskIds = taskSummaryList.stream().map(TaskSummary::getId).toList();
+
+    if (!taskSummaryList.isEmpty()) {
+      taskMapper.updateTerminated(taskIds, taskSummaryList.get(0));
```
CRoberto1926, line 2468: Why do you take the first task of the list and use it as referencedTask in the UPDATE query?
I would rather have modified and completed explicitly set as `now` via parameter. Also, `state` could be either a parameter or a constant in the query.

My point is: I think is confusing to pass a "referenceTask" when you actually need only the "time of termination" timestamp.

lib/kadai-core/src/main/java/io/kadai/task/internal/TaskMapper.java
```diff
@@ -195,6 +195,14 @@ void updateTransfered(
   void updateCompleted(
       @Param("taskIds") List<String> taskIds, @Param("referenceTask") TaskSummary referenceTask);
 
+  @Update(
+      "<script>"
+          + " UPDATE TASK SET COMPLETED = #{referenceTask.completed}, MODIFIED = #{referenceTask.modified}, STATE = #{referenceTask.state}"
+          + " WHERE ID IN <foreach item='taskId' index='index' separator=',' open='(' close=')' collection='taskIds'>#{taskId}</foreach>"
+          + "</script>")
+  void updateTerminated(
+      @Param("taskIds") List<String> taskIds, @Param("referenceTask") TaskSummary referenceTask);
```
CRoberto1926, line 204: Opinion on "referenceTask" on the previous comment :)

lib/kadai-core/src/main/java/io/kadai/task/internal/TaskServiceImpl.java
```diff
@@ -2412,6 +2418,74 @@ private void updateTasksToBeCompleted(Stream<TaskSummaryImpl> taskSummaries, Ins
     }
   }
 
+  private BulkOperationResults<String, KadaiException> terminateTasks(List<String> taskIds)
+      throws InvalidArgumentException, NotAuthorizedException {
+    kadaiEngine.getEngine().checkRoleMembership(KadaiRole.ADMIN, KadaiRole.TASK_ADMIN);
+    try {
+      kadaiEngine.openConnection();
+      if (taskIds == null) {
+        throw new InvalidArgumentException("TaskIds can't be used as NULL-Parameter.");
```
CRoberto1926: Maybe a couple of tests for `taskIds == null` and `taskIds.isEmpty()`?

lib/kadai-core/src/main/java/io/kadai/task/internal/TaskServiceImpl.java
bruderj15: Completely on your side. However I went with what we were already doing for similar methods:

https://github.com/kadai-io/kadai/blob/89f6b57b3b6a88db2dc29225f6c476ff4f7c197b/lib/kadai-core/src/main/java/io/kadai/task/internal/TaskMapper.java#L190-L212

I'd keep it that way.
If it bothers you too much, consider opening a separate issue :)

CRoberto1926: It's fine, I had already seen the pattern, but I wanted to leave a comment nevertheless :D

CRoberto1926 (approved):

PR #1258, opened 2025-07-23
Closes #1208 - Make Auditing Proxy-able

bruderj15:
## Definition of Done

- [x] The corresponding ticket is in state `In Review`
- [x] The corresponding ticket is attached to this Pull-Request
- [x] The commit-message-format `Closes #ISSUE_ID - PROBLEM/SOLUTION` 
  - is present as single-commit already or
  - will be squashed on merge
- [x] The changes pass the SonarQubeCloud-Quality-Gate
- [x] If the documentation needs an update, following there's a link to it:
- [x] If extra release-notes are required, they're listed indented below:
  - Added `KadaiEngine::<T>runAs(Supplier<T> action,  KadaiRole proxy, String userId)` and overloaded siblings
  - You can now run actions with the permissions of a specific role for a certain user. This can be thought of temporarily lifting the user to the specified role. The role is therefore acting as proxy. 
  - There still remains the previously existing `KadaiEngine::<T>runAsAdmin(Supplier<T> action)`. 
    - The key difference is in the downstream creation of Kadai History-Events. 
    - The newly added methods log the `userId` for the lifted user **and** the `accessId` of the proxy-role. 
    - The legacy-method **only** logs an `accessId` of `KadaiRole::ADMIN` as the executing user. 
    There is no change in behavior here.
  - ‼️Renamed `CurrentUserContext::getUserid` to `CurrentUserContext::getUserId`
  - ‼️Removed support for `WebSphere`-Authentication

PR #1789, opened 2026-03-17
Closes #1279 - Centralize Auditing via `KadaiEventBus`

### LINKED ISSUE #1279: Event-Management should move auditing-logic from Domain-Services to internal orchestrator, e.g.`HistoryEventManager`
Related to #545:

## What's wrong?

#545 already points out unfortunate design-decisions for the current state of Kadai's Event-Management.
I can add further to this.

Afaik we only construct internal events in the Service-Layer.
There it always looks like this, e.g. [`TaskServiceImpl::reopen`](https://github.com/kadai-io/kadai/blob/1c0e5902faa2fdd8588ff26f3a8bb0db1fd732bb/lib/kadai-core/src/main/java/io/kadai/task/internal/TaskServiceImpl.java#L701):

```java
public Task reopen(String taskId) throws ... {
  // Some Service-Work...
  // ...
  String userId = kadaiEngine.getEngine().getCurrentUserContext().getUserId();
  if (historyEventManager.isEnabled()) {
    String changeDetails =
        ObjectAttributeChangeDetector.determineChangesInAttributes(oldTask, task);
    historyEventManager.createEvent(
        new TaskReopenedEvent(
            IdGenerator.generateWithPrefix(IdGenerator.ID_PREFIX_TASK_HISTORY_EVENT),
            task,
            userId,
            changeDetails));
  }
}
```

Not so bad, isn't it?
Wrong, **very bad**.

## Why is it wrong?

While working on #1208 I discovered that adding a field to every event-type is tiresome, especially adjusting constructed values (events).
It get's worse.
The logic to retrieve the in #1208 required `proxyAccessId`, which we there want to log, is always identical.
This is similarly already visible in the snippet above - for the `userId`.
And it's no wonder.

The current design does **not separate concerns** here.
The service-layer - ought to orchestrate - does authentification (entirely).
While this _may_ be fine for domain-logic, it certainly is not for auditing.

## How to get it right?

Instead some kind of mediator (terminology from Event-Driven-Architectures) should add the `userId` and newly required `proxyAccessId`.
In the implementation for #545 suggested with #994, the `KadaiEventBroker` could be extended to a `KadaiEventBus` to decorate all events.
Yet, my preferred way would be another mediator in front of the `KadaiEventPublisher` - explicitly for adding such ids.
The latter would also work in asynchronous contexts. 
As they are highly likely not be required async though, the approach with the bus should be favored.

bruderj15:
## Definition of Done

  - will be squashed on merge


lib/kadai-core/src/main/java/io/kadai/common/internal/InternalKadaiEngine.java
```diff
@@ -107,7 +107,7 @@ default void executeInDatabaseConnection(Runnable runnable) {
    *
    * @return the HistoryEventProducer instance.
    */
-  KadaiEventBroker getKadaiEventBroker();
+  KadaiEventBus getKadaiEventBroker();
```
CRoberto1926: `getKadaiEventBus()`?

lib/kadai-core/src/main/java/io/kadai/common/internal/KadaiEngineImpl.java
```diff
@@ -609,8 +609,8 @@ public KadaiEngine getEngine() {
     }
 
     @Override
-    public KadaiEventBroker getKadaiEventBroker() {
-      return kadaiEventBroker;
+    public KadaiEventBus getKadaiEventBroker() {
```
CRoberto1926: `getKadaiEventBus()`?

lib/kadai-core/src/test/java/acceptance/history/KadaiEventBusTest.java
```diff
@@ -45,127 +47,138 @@ void testKadaiEventBrokerIsNotEnabled() {
...
 
+  @Test
+  void should_EnrichUserAndProxyAccessId_For_Dispatch() {
+    final KadaiEventBus eventBus = new KadaiEventBus(kadaiEngine);
+    final KadaiEvent event = new TaskHistoryEvent();
+
+    kadaiEngine.runAs(() -> eventBus.dispatch(event), KadaiRole.ADMIN, "roberto");
+
+    assertThat(event.getUserId()).isEqualTo("roberto");
+    assertThat(event.getProxyAccessId()).isEqualTo("uid=admin,cn=users,ou=test,o=kadai");
```
CRoberto1926, line 179: What exactly is a "proxyAccessId" and how is it defined?

bruderj15: It's an access-id that has been supplied by some outer context to execute the contained context as a user.
It was introduced with #1258 upon a feature-request.

That way one can run an action for a user but _proxy_ his permissions by temporarily extending the user-context to include said access-id.

CRoberto1926 (approved):
LGTM

PR #1801, opened 2026-03-20
Closes #1779 - Add 'CreateTaskPostprocessor' SPI

### LINKED ISSUE #1779: Implement CreateTaskPostProcessor SPI for auto-claim after task creation
When a user task is created from a process (e.g., via Camunda), there is currently no mechanism to automatically claim it or execute any post-creation logic. This prevents seamless automation workflows where tasks should be claimed immediately upon creation. We need a CreateTaskPostProcessor SPI that provides a standardized extension point for executing custom logic after a task is created. 

**Implementation:**
A new CreateTaskPostProcessor interface will be introduced with the following method:

`Task processTaskAfterCreation(Task taskToProcess);`

**ToDo's:**

- [ ]  Define the SPI interface
- [ ]  Provide example implementation
- [ ]  Write tests
- [ ]  Update documentation

lucalimbach:
## Definition of Done

  - will be squashed on merge


lib/kadai-core/src/main/java/io/kadai/task/internal/TaskServiceImpl.java
```diff
@@ -417,10 +420,10 @@ public Task createTask(Task taskToCreate)
 
       createTaskCreatedHistoryEvent(task);
 
-      return task;
     } finally {
       kadaiEngine.returnConnection();
     }
+    return postprocessTaskCreation(task);
```
bruderj15: I think we want to keep it inside the try-block so we're guaranteed to be in the same transaction, right?
So I'd move `return postprocessTaskCreation(task)` back into the try-block where `return task` originally was.

What do you think?

lib/kadai-core/src/main/java/io/kadai/task/internal/TaskServiceImpl.java
```diff
@@ -1724,6 +1723,18 @@ private TaskImpl preprocessTaskCreation(Task taskToCreate) {
...
+      task = (TaskImpl) createTaskPostProcessorManager.processTaskAfterCreation(task);
+    }
+
+    if (LOGGER.isDebugEnabled()) {
+      LOGGER.debug("Method postprocessTaskCreation() post-processed Task '{}'.", task.getId());
+    }
+
+    return task;
+  }
+
```
bruderj15:
```suggestion
  private Task postprocessTaskCreation(TaskImpl task) {
    Task task = task;
    if (createTaskPostProcessorManager.isEnabled()) {
      task = createTaskPostProcessorManager.processTaskAfterCreation(task);
    }

    if (LOGGER.isDebugEnabled()) {
      LOGGER.debug("Method postprocessTaskCreation() post-processed Task '{}'.", task.getId());
    }

    return task;
  }

```

Unnecessary strict type here afaics.

lib/kadai-core-test/src/test/java/acceptance/taskpostprocessing/CreateTaskPostProcessingAccTest.java
```diff
@@ -0,0 +1,119 @@
...
+import io.kadai.workbasket.api.models.WorkbasketSummary;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+
+/** Acceptance test for "task post-processing" scenario. */
+@KadaiIntegrationTest
+@WithServiceProvider(
+        serviceProviderInterface = CreateTaskPostprocessor.class,
+        serviceProviders = TestCreateTaskPostProcessorProvider.class)
+class CreateTaskPostProcessingAccTest {
```
bruderj15, line 45: Can you add a test highlighting the transactional boundaries on failure?

bruderj15 (commented):
Looks good. Few minor comments attached :)

lib/kadai-core/src/main/java/io/kadai/task/internal/TaskServiceImpl.java
lucalimbach: The post-processor is intentionally placed outside the try block. 
Inside the try block (a single transaction):
- Post-processor fails → The task creation is rolled back
- Task creation fails → The post-processor is never executed

Outside the try block (two separate transactions):
- Post-processor fails → The task is already committed to the database
- Task creation fails → The post-processor is never executed

I thought it would be better to ensure the task is created before any post-processing logic is executed, regardless of whether the post-processor succeeds or fails.
What do you think, should a failed post-processor 
roll back the task creation?

lib/kadai-core/src/main/java/io/kadai/spi/task/api/CreateTaskPostprocessor.java
```diff
@@ -0,0 +1,28 @@
...
+ * The CreateTaskPostProcessor allows to implement customized behavior after the given {@linkplain
+ * Task} has been created.
+ */
+public interface CreateTaskPostprocessor {
+  /**
+   * Perform any action after a {@linkplain Task} has been created through {@linkplain
+   * TaskService#createTask(Task)}.
+   *
+   * <p>This SPI is executed within the same transaction staple as {@linkplain
+   * TaskService#createTask(Task)}.
```
bruderj15, line 18: Here it is

lib/kadai-core/src/main/java/io/kadai/task/internal/TaskServiceImpl.java
bruderj15: ~I think separate transactions are correct~. I was mislead as somewhere in your JavaDocs (probably on the service-fun) it says _same transactions staple_. Commented on this below.

bruderj15: @lucalimbach I checked the existing SPIs for their behavior. They have the action and triggered post-processing SPI all run within the same transaction:

https://github.com/kadai-io/kadai/blob/f25a5ba6e253c085b28d311b1dc7946e7acb86b9/lib/kadai-core/src/main/java/io/kadai/task/internal/TaskServiceImpl.java#L2034-L2039

https://github.com/kadai-io/kadai/blob/f25a5ba6e253c085b28d311b1dc7946e7acb86b9/lib/kadai-core/src/main/java/io/kadai/task/internal/TaskServiceImpl.java#L2085-L2090

We should therefore here stay within one transaction.

bruderj15 (commented):
CI is failing due to an arch violation.
Looks good otherwise.

bruderj15 (approved):

PR #1884, opened 2026-04-27
Closes #1883 - Reduced payloads in `TaskCleanupJob`

### LINKED ISSUE #1883: TaskCleanupJob to load task-id and parent-business-process-id in batches instead of all tasks

bruderj15:
## Definition of Done

  - will be squashed on merge


lib/kadai-core/src/main/java/io/kadai/task/internal/models/TaskCleanupSummary.java
```diff
@@ -0,0 +1,76 @@
...
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ *
+ *
+ */
+
+package io.kadai.task.internal.models;
```
gitgoodjhe: I'm not sure if this is the right place for this model since we have a separate "jobs" package and the model is only used in job contexts, right?

lib/kadai-core/src/main/java/io/kadai/task/internal/TaskMapper.java
```diff
@@ -332,4 +333,21 @@ List<Pair<String, String>> getTaskAndWorkbasketIdsNotAuthorizedFor(
...
   void decrementNumberOfComments(@Param("id") String id, @Param("modified") Instant modified);
+
+  @Select(
+      "<script>SELECT ID, PARENT_BUSINESS_PROCESS_ID FROM TASK "
+          + "WHERE COMPLETED &lt;= #{completedBefore} "
+          + "<if test=\"_databaseId == 'db2'\">with UR </if>"
+          + "</script>")
+  @Result(property = "taskId", column = "ID")
+  @Result(property = "parentBusinessProcessId", column = "PARENT_BUSINESS_PROCESS_ID")
+  List<TaskCleanupSummary> findTasksCompletedBefore(
```
gitgoodjhe, line 344: are we missing tests?

lib/kadai-core/src/main/java/io/kadai/common/internal/KadaiEngineImpl.java
```diff
@@ -293,6 +293,13 @@ public PriorityServiceManager getPriorityServiceManager() {
     return priorityServiceManager;
   }
 
+  // This should be part of the InternalKadaiEngine. Unfortunately the jobs don't have access to
+  // that engine.
+  // Therefore, this getter exits and will be removed as soon as our jobs will be refactored.
```
gitgoodjhe:
```suggestion
  // Therefore, this getter exists and will be removed as soon as our jobs will be refactored.
```

lib/kadai-core/src/main/java/io/kadai/common/internal/KadaiEngineImpl.java
```diff
@@ -293,6 +293,13 @@ public PriorityServiceManager getPriorityServiceManager() {
...
+  public TaskMapper getTaskMapper() {
```
gitgoodjhe, line 299: On first glance I don't like this

lib/kadai-core/src/main/java/io/kadai/task/internal/TaskMapper.java
```diff
@@ -332,4 +333,21 @@ List<Pair<String, String>> getTaskAndWorkbasketIdsNotAuthorizedFor(
...
+      @Param("completedBefore") Instant completedBefore);
+
+  @Select(
+      "<script>SELECT COUNT(*) FROM TASK "
+          + "WHERE PARENT_BUSINESS_PROCESS_ID = #{parentBpi} "
+          + "<if test=\"_databaseId == 'db2'\">with UR </if>"
+          + "</script>")
+  long countTasksByParentBusinessProcessId(@Param("parentBpi") String parentBpi);
```
gitgoodjhe: An idea I had when reading through the code, just as a thought: This means that count method will be called for each parent business id, which can exceed multiple hundred thousands -> hundred thousands of (granted, cheap) database calls. This was already the case before the changes, just not directly in mapper but through query.

 Maybe we have a huge opportunity to optimize further here, for example by something like this:

SELECT PARENT_BUSINESS_PROCESS_ID, COUNT(*)
FROM TASK
WHERE PARENT_BUSINESS_PROCESS_ID IN (...)
GROUP BY PARENT_BUSINESS_PROCESS_ID

-> 1 call (per chunk), then we can compare them with the ones in memory etc
Ofc this would require chunks since hundreds of thousands of IN paramters would probably mess up the database. Even if we made the chunks "just" a size of 100 ids, then we would still be getting 100 times less database calls.

Maybe you can think about this and tell me if i overlook something here, since I don't have the time to think it through fully

gitgoodjhe (changes requested):

lib/kadai-core/src/main/java/io/kadai/common/internal/KadaiEngineImpl.java
bruderj15: Neither do I honestly.
I went with the same approach as a few lines upwards with `PriorityServiceManager`.
No straightforward solution here but okayish for the `KadaiEngineImpl` to expose for now imo.
I openend a new ticket with #1889 and would keep this approach here as is.

lib/kadai-core/src/main/java/io/kadai/task/internal/TaskMapper.java
bruderj15: No. Although we don't directly test the mappers, the code is covered with already existing tests in `TaskCleanupJobAccTest`. 
SonarQube coverage confirms this.

lib/kadai-core/src/main/java/io/kadai/task/internal/TaskMapper.java
bruderj15: Went with:
- finding all eligible parent-business-process-ids that may have any task to clean up
- anti-joining them on tasks violating the parent-business-process-id constraint
- fetching all tasks

Here's the query in a more readable form than from the MyBatis snippet:
```sql
WITH ELIGIBLE_PARENT_BUSINESS_PROCESS_IDS AS (
  SELECT candidate.PARENT_BUSINESS_PROCESS_ID
  FROM (
    SELECT DISTINCT PARENT_BUSINESS_PROCESS_ID
    FROM TASK
    WHERE COMPLETED <= ?
      AND PARENT_BUSINESS_PROCESS_ID IS NOT NULL
      AND PARENT_BUSINESS_PROCESS_ID <> ''
  ) candidate
  LEFT JOIN TASK blocker
    ON blocker.PARENT_BUSINESS_PROCESS_ID = candidate.PARENT_BUSINESS_PROCESS_ID
   AND (blocker.COMPLETED IS NULL OR blocker.COMPLETED > ?)
  WHERE blocker.ID IS NULL
)
SELECT t.ID, t.PARENT_BUSINESS_PROCESS_ID
FROM TASK t
LEFT JOIN ELIGIBLE_PARENT_BUSINESS_PROCESS_IDS eligible
  ON eligible.PARENT_BUSINESS_PROCESS_ID = t.PARENT_BUSINESS_PROCESS_ID
WHERE t.COMPLETED <= ?
  AND (
    t.PARENT_BUSINESS_PROCESS_ID IS NULL
    OR t.PARENT_BUSINESS_PROCESS_ID = ''
    OR eligible.PARENT_BUSINESS_PROCESS_ID IS NOT NULL
  )
```

Also added (optional but) recommended indices on Task (parent-business-process-id, completed) and its reverse.

I am quite confident this is the most efficient approach.

gitgoodjhe (approved):

PR #1939, opened 2026-05-26
Closes #1915 - Fix integration tests don't fail missing connection & `TaskCleanupJob` conn

bruderj15:
## Definition of Done

- [x] The corresponding ticket is in state `In Review`
- [x] The corresponding ticket is attached to this Pull-Request
- [x] The commit-message-format `Closes #ISSUE_ID - PROBLEM/SOLUTION` 
  - is present as single-commit already or
  - will be squashed on merge
- [x] The changes pass the SonarQubeCloud-Quality-Gate
- [x] If the documentation needs an update, following there's a link to it:
- [x] If extra release-notes are required, they're listed indented below:

PR #1966, opened 2026-06-10
Closes #1883/#1915 - Backport missing Database Connection in and optimize `TaskCleanupJob`

### LINKED ISSUE #1883: TaskCleanupJob to load task-id and parent-business-process-id in batches instead of all tasks

### LINKED ISSUE #1915: Integration Tests do not fail when missing to open connection and setting schema

bruderj15:
Cherry-picked commits of related upstream PRs and aligned them.

## Definition of Done

  - will be squashed on merge


lib/kadai-core/src/main/java/io/kadai/task/internal/jobs/TaskCleanupJob.java
```diff
@@ -89,55 +77,17 @@ protected String getType() {
...
-          tasksToDelete.stream()
-              .filter(
-                  taskSummary ->
-                      !taskIdsNotAllCompletedSameParentBusiness.contains(
-                          taskSummary.getParentBusinessProcessId()))
-              .toList();
-    }
-
-    return tasksToDelete;
+  private List<TaskCleanupSummary> getTasksCompletedBefore(Instant untilDate) {
```
gitgoodjhe: why don't we define an api in the TaskService and use that instead of making kadaiEngine able to do databaseTransactions? This kinda violates our api design I think

lib/kadai-core/src/main/java/io/kadai/task/internal/TaskMapper.java
```diff
@@ -341,4 +342,44 @@ List<Pair<String, String>> getTaskAndWorkbasketIdsNotAuthorizedFor(
...
+      "<script>SELECT ID, PARENT_BUSINESS_PROCESS_ID FROM TASK "
+          + "WHERE COMPLETED &lt;= #{completedBefore} "
+          + "<if test=\"_databaseId == 'db2'\">with UR </if>"
+          + "</script>")
+  @Result(property = "taskId", column = "ID")
+  @Result(property = "parentBusinessProcessId", column = "PARENT_BUSINESS_PROCESS_ID")
+  List<TaskCleanupSummary> findTasksCompletedBefore(
+      @Param("completedBefore") Instant completedBefore);
+
+  @Select(
```
gitgoodjhe, line 356: did you test this manually once? Last time the tests didn't catch the problem

gitgoodjhe (changes requested):

lib/kadai-core/src/main/java/io/kadai/task/internal/jobs/TaskCleanupJob.java
bruderj15: Valid. Went with it to keep task-logic and type-usage (`TaskCleanupSummary`) job-local.

Didn't move it into the API now but into `TaskServiceImpl` which is fine (also solves the mapper access problem).

lib/kadai-core/src/main/java/io/kadai/task/internal/TaskMapper.java
bruderj15: I didn't. These are cherry-picks from #1939 which (introduces this exact fix here and) enforces schema respect.

NB: This PR here just backports for a 11.4.1 patch.

gitgoodjhe (approved):

PR #2045, opened 2026-07-21
Closes #2044 - Extend Task with `creatorLongName`

### LINKED ISSUE #2044: Extend `TaskSummary` with `creatorLongName`

bruderj15:
## Definition of Done

  - will be squashed on merge


lib/kadai-core/src/main/java/io/kadai/task/internal/TaskQueryImpl.java
```diff
@@ -87,6 +87,7 @@ public class TaskQueryImpl implements TaskQuery {
   private boolean addAttachmentClassificationNameToSelectClauseForOrdering = false;
   private boolean addWorkbasketNameToSelectClauseForOrdering = false;
   private boolean joinWithUserInfo;
+  private boolean joinWithCreatorUserInfo;
```
leantios, line 90: Why do we need a separate field here and can't use `joinWithUserInfo`?

leantios (commented):
Looks really good besides one question!

lib/kadai-core/src/main/java/io/kadai/task/internal/TaskQueryImpl.java
bruderj15: Very valid question.

In the MyBatis SQL we do:

```
...
"<if test=\"joinWithUserInfo\">, owner_info.LONG_NAME AS OWNER_LONG_NAME</if>"
"<if test=\"joinWithCreatorUserInfo\">, creator_info.LONG_NAME AS CREATOR_LONG_NAME</if>"
...
```

One may get away with dynamically selecting column-names but that get's awkward down the line for usages of  them in the query.

leantios (approved):
Good Job!

PR #2050, opened 2026-07-22
Closes #2049 - migrate `kadai.properties` from custom-parsed to Spring-parsed

### LINKED ISSUE #2049: Migrate `KadaiConfiguration` with `kadai.properties` to `@ConfigurationProperties`

bruderj15:
- Replaced custom parsing with spring-boot binding. Therefore `kadai.properties` is now
  - much better represented in-code with dedicated classes instead of dangling fields
  - fully type-safe
  - helping end-users when configuring by auto-completing valid properties and values
- Moved defaults from hidden internal `KadaiConfiguration.Builder` to publicly facing `KadaiConfiguration`
- Replace custom validation with jakarta for spec and hibernate for impl to move constraints away from black hole to the actual config-fields

Spring Boot is only used for the binding (we do not use a starter-dep).
Therefore Kadai still runs with just plain Java.

**I think we should ship a migration-script for the properties**. I would add that once we approve the direction suggested in this PR.

## Definition of Done

  - will be squashed on merge
- [x] If the documentation needs an update, following there's a link to it: after approval
- [x] If extra release-notes are required, they're listed indented below: after approval


bruderj15:
We cannot ship spring-boot into the core - pulls too many transitive dependencies.

PR #2067, opened 2026-07-28
Closes #2066 - Move supported Postgres to 17

### LINKED ISSUE #2066: Move supported Postgres to 17

bruderj15:
## Definition of Done

  - will be squashed on merge
- [x] If the documentation needs an update, following there's a link to it: https://github.com/kadai-io/kadai-doc/pull/238

### DIFF
--- a/lib/kadai-core/src/main/java/io/kadai/monitor/internal/MonitorMapperSqlProvider.java
+++ b/lib/kadai-core/src/main/java/io/kadai/monitor/internal/MonitorMapperSqlProvider.java
@@ -53,13 +53,14 @@ public class MonitorMapperSqlProvider {
         + OPENING_WHERE_TAG
         + taskWhereStatements()
         + "<if test=\"report.combinedClassificationFilter != null\">"
-        + "AND <foreach collection='report.combinedClassificationFilter' "
-        + "item='item' separator='OR'> "
-        + "T.CLASSIFICATION_ID = #{item.taskClassificationId}"
+        + "AND (<foreach collection='report.combinedClassificationFilter' "
+        + "item='item' separator=' OR '> "
+        + "(T.CLASSIFICATION_ID = #{item.taskClassificationId} "
         + "<if test=\"item.attachmentClassificationId != null\">"
-        + "AND A.CLASSIFICATION_ID = #{item.attachmentClassificationId}"
+        + "AND A.CLASSIFICATION_ID = #{item.attachmentClassificationId} "
         + "</if>"
-        + "</foreach>"
+        + ") "
+        + "</foreach>) "
         + "</if>"
         + "AND T.${timestamp} IS NOT NULL "
         + CLOSING_WHERE_TAG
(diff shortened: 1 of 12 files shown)


lib/kadai-core/src/main/java/io/kadai/monitor/internal/MonitorMapperSqlProvider.java
```diff
@@ -53,13 +53,14 @@ public static String getTaskCountOfWorkbaskets() {
         + OPENING_WHERE_TAG
         + taskWhereStatements()
         + "<if test=\"report.combinedClassificationFilter != null\">"
-        + "AND <foreach collection='report.combinedClassificationFilter' "
-        + "item='item' separator='OR'> "
-        + "T.CLASSIFICATION_ID = #{item.taskClassificationId}"
+        + "AND (<foreach collection='report.combinedClassificationFilter' "
```
leantios, line 56: Why do we need these changes?

leantios (commented):
I don' know how relevant this is but in copilot-instructions.md, we mention v14.
Also, the Dockerfile in docker-databases/postgres_14 should be updated, no?
As well as the compose.yml for the GMT.

lib/kadai-core/src/main/java/io/kadai/monitor/internal/MonitorMapperSqlProvider.java
bruderj15: Because Postgres complained - seems like they've gotten more strict.

The space between `?` and `AND`/`OR` are needed as described in the [spec](https://www.postgresql.org/docs/17/sql-expressions.html#SQL-SYNTAX-PARAMETERS). It would otherwise render
`T.CLASSIFICATION_ID = $2AND A.CLASSIFICATION_ID = $3OR T.CLASSIFICATION_ID = $4`, which Postgres now seems to complain about.
You can see the actual error in [CI](https://github.com/kadai-io/kadai/actions/runs/30360588937/job/90282362820#step:8:619).

Regarding the extra parentheses: I figured they were missing when inspecting why above error occurred.
We'd previously have MyBatis render:
```sql
WHERE <report filters>
  AND classification = c1 AND attachment = a1
   OR classification = c2 AND attachment = a2
   OR classification = c3
  AND due IS NOT NULL
```

although - due to AND < OR preference - we actually want
```sql
WHERE <report filters>
  AND (
    classification = c1 AND attachment = a1
    OR classification = c2 AND attachment = a2
    OR classification = c3
  )
  AND due IS NOT NULL
```

bruderj15:
> I don' know how relevant this is but in copilot-instructions.md, we mention v14.
Also, the Dockerfile in docker-databases/postgres_14 should be updated, no?

Thank you. Bumped.

> As well as the compose.yml for the GMT.

Scope of new issue: https://github.com/kadai-io/kadai/issues/2094

lib/kadai-core/src/main/java/io/kadai/monitor/internal/MonitorMapperSqlProvider.java
leantios: Thanks for the detailed explanations! Funny that the error complains about "Müll" :D 
Do we need a test for the found bug?

leantios (commented):
The Dockerfile in docker-databases is still missing

PR #2069, opened 2026-07-29
Closes #2068 - Optimize general Task Bulk-Delete & `TaskCleanupJob` performance

### LINKED ISSUE #2068: Optimize `TaskCleanupJob` performance

bruderj15:
- Heavy hitter was the eligibility query
  - 100k tasks
  - 1:10 pbp:task
  - Postgres chose: `Nested Loop with Rows Removed by Join Filter: 999900000` (ouch)
- `TaskServiceImpl::deleteTasks` checked task-invariants in `O(n²)` due to `ArrayList`, now in `O(n)` with `HashMap`

Extra finding: old impl didn't fully respect (at least renew) the job-lock.

---

New SQL without the MyBatis pain:
```sql
WITH tasks_with_parent_completion AS (
    SELECT
        id,
        completed,
        COUNT(*) OVER (
            PARTITION BY parent_business_process_id
        ) AS parent_task_count,
        COUNT(completed) OVER (
            PARTITION BY parent_business_process_id
        ) AS parent_completed_task_count,
        MAX(completed) OVER (
            PARTITION BY parent_business_process_id
        ) AS parent_max_completed
    FROM task
    WHERE parent_business_process_id IS NOT NULL
      AND CHARACTER_LENGTH(parent_business_process_id) > 0
)

SELECT id
FROM tasks_with_parent_completion
WHERE completed <= :completed_before
  AND parent_task_count = parent_completed_task_count
  AND parent_max_completed <= :completed_before

UNION ALL

SELECT id
FROM task
WHERE completed <= :completed_before
  AND (
      parent_business_process_id IS NULL
      OR CHARACTER_LENGTH(parent_business_process_id) = 0
  );
```

---

Ran a small benchmark with 
- Postgres 
- Cleanup-batch-size=5k 
- 1:10 pbp:task

| Tasks | Historical baseline | Final runtime | Improvement |
| ---: | ---: | ---: | ---: |
| 20,000 | 5.887s | 1.143s | 5.15× faster, 80.6% less runtime |
| 100,000 | 25.777s | 5.447s | 4.73× faster, 78.9% less runtime |
| 1,500,000 | >17 minutes, cancelled during selection | 96.681s | >10.55× faster, >90.5% less runtime |

## Definition of Done

  - will be squashed on merge
- [x] If the documentation needs an update, following there's a link to it: [upon approval of approach](https://github.com/kadai-io/kadai-doc/pull/244)


lib/kadai-core/src/main/java/io/kadai/common/api/KadaiEngine.java
```diff
@@ -37,7 +37,7 @@
 
 /** The KadaiEngine represents an overall set of all needed services. */
 public interface KadaiEngine {
-  String MINIMAL_KADAI_SCHEMA_VERSION = "12.0.0";
+  String MINIMAL_KADAI_SCHEMA_VERSION = "12.1.0";
```
bruderj15: Debatable.

We might just keep minimal schema version at 12.0.0 and recommend the extra index by keeping the migration script.

Up to you.

lib/kadai-core/src/main/java/io/kadai/spi/history/api/KadaiEventConsumer.java
```diff
@@ -40,6 +40,9 @@ public interface KadaiEventConsumer<T extends KadaiEvent> extends KadaiInitializ
   /**
    * Consumes multiple events without guaranteeing order.
    *
+   * <p>This method does not opt a consumer into batched event-bus dispatch. Implement {@link
+   * BatchKadaiEventConsumer} for that purpose.
+   *
    * @param events the events to consume
    */
   default void consumeAll(Collection<T> events) {
```
bruderj15, line 48: I don't like this myself.
Deprecating it breaks compile though.

lib/kadai-core/src/main/java/io/kadai/common/internal/jobs/JobRunner.java
```diff
@@ -46,38 +47,55 @@ public void registerTransactionProvider(KadaiTransactionProvider txProvider) {
...
+    if (job.isAsync()) {
+      job.scheduleNextJob();
     }
+    jobService.deleteJob(scheduledJob);
   }
 
   private ScheduledJob lockJob(ScheduledJob job) {
     String hostAddress = getHostAddress();
-    String owner = hostAddress + " - " + Thread.currentThread().getName();
+    String owner = hostAddress + " - " + UUID.randomUUID();
```
bruderj15, line 98: More robust for multi-instance.
Host-Address could be removed now but kept it to keep the information (though I doubt anyone cares).

lib/kadai-core/src/main/java/io/kadai/task/internal/TaskMapper.java
```diff
@@ -344,42 +343,38 @@ List<Pair<String, String>> getTaskAndWorkbasketIdsNotAuthorizedFor(
...
-      @Param("completedBefore") Instant completedBefore);
+  List<String> findTasksCompletedBefore(@Param("completedBefore") Instant completedBefore);
 
   @Select(
       "<script>"
-          + "WITH ELIGIBLE_PARENT_BUSINESS_PROCESS_IDS AS ("
-          + "SELECT candidate.PARENT_BUSINESS_PROCESS_ID "
-          + "FROM ("
-          + "SELECT DISTINCT PARENT_BUSINESS_PROCESS_ID "
+          + "WITH TASKS_WITH_PARENT_COMPLETION AS ("
```
gitgoodjhe, line 354: I guess this is thouroughly tested right?

gitgoodjhe:
I like theses changes! While skipping through the logic of the classes one thing came to mind without checking thouroughly: 

when we delete tasks, we also delete attachments and objectReferences before deleting the actual tasks. Now let's say we deleted both successfully but the task deletion fails. Then this code will swallow the exception, no?

 private int deleteTasksTransactionally(List<String> tasksToBeDeleted) {
    return KadaiTransactionProvider.executeInTransactionIfPossible(
        txProvider,
        () -> {
          int deletedTasks;
          try {
            deletedTasks = deleteTasks(tasksToBeDeleted);
          } catch (Exception ex) {
            LOGGER.warn("Could not delete tasks.", ex);
            return 0;
          }
          renewLock();
          return deletedTasks;
        });
  }
  
  -> are we running the risk of introducing inconsistent data here? If yes, this is not introduced in this PR, it was already like this.

lib/kadai-core/src/main/java/io/kadai/common/api/KadaiEngine.java
bruderj15: Keeping 12.0.0 instead and have migration-script be optional

bruderj15:
>  Then this code will swallow the exception, no?

The batch fails and gets logged but is not swallowed as subsequent job-executions will read them as eligible and then attempt to delete again.
The logged exception is the part in `TaskServiceImpl::deleteTasks` failing (e.g. deleting attachments for a specific task).

> are we running the risk of introducing inconsistent data here?

No. The entire batch runs in one transaction.
Therefore, also a single task and all its relatives are deleted in exactly one transaction (its batches' one).
Corrupt data here would only be possible if `TaskServiceImpl::deleteTasks` span multiple transactions per task.

This being fairly obvious, did you mean something else I am missing?

gitgoodjhe:
> > Then this code will swallow the exception, no?
> 
> The batch fails and gets logged but is not swallowed as subsequent job-executions will read them as eligible and then attempt to delete again. The logged exception is the part in `TaskServiceImpl::deleteTasks` failing (e.g. deleting attachments for a specific task).
> 
> > are we running the risk of introducing inconsistent data here?
> 
> No. The entire batch runs in one transaction. Therefore, also a single task and all its relatives are deleted in exactly one transaction (its batches' one). Corrupt data here would only be possible if `TaskServiceImpl::deleteTasks` span multiple transactions per task.
> 
> This being fairly obvious, did you mean something else I am missing?

I don't have Access to my IDE currently, so just to make sure:

something like this can not ever happen?

DELETE attachments       ✓
DELETE object refs        ✓
DELETE task               ✓
publish TaskDeletedEvent  ✗

deleteTasks()
   ↓
event publishing throws
   ↓
TaskCleanupJob catches it
   ↓
returns 0 normally
   ↓
outer transaction does NOT see exception

-> gets committed in the end. Tasks don't exist anymore in the next run, so events will never be written

bruderj15:
> something like this can not ever happen?

You were right. It could very well happen because any exception was caught inside the `KadaiTransactionProvider::executeInTransactionIfPossible` lambda in `TaskCleanupJob`.

I moved it outside the lambda and added an acceptance-test proving full batch rollback.

bruderj15:
Given your otherwise positive feedback I will:
- merge this
- because this is a blocker
- but not release it until you're back @gitgoodjhe and we agreed on it

PR #2120, opened 2026-08-13
Closes #2013 - Remove setTimeout()-calls from unit-tests 

### LINKED ISSUE #2013: Analysis - Removal of setTimeout() calls from unit tests
In many tests setTimeout() is used together with Promises to just wait, think of Thread.sleep().

We should analyze if it possible to remove these calls and use better approaches in the unit tests.

SanzharBerdikozhoev:
## Definition of Done

  - will be squashed on merge

### DIFF
--- a/web/src/app/administration/components/workbasket-information/workbasket-information.component.spec.ts
+++ b/web/src/app/administration/components/workbasket-information/workbasket-information.component.spec.ts
@@ -484,8 +484,9 @@ describe('WorkbasketInformationComponent', () => {
     const formsValidatorService = TestBed.inject(FormsValidatorService);
     (formsValidatorService.validateFormInformation as any).mockReturnValueOnce(Promise.resolve(false));
     await component.onSubmit();
-    await new Promise((resolve) => setTimeout(resolve, 0));
-    expect(showErrorSpy).toHaveBeenCalledWith('WORKBASKET_SAVE');
+    vi.waitFor(() => {
+      expect(showErrorSpy).toHaveBeenCalledWith('WORKBASKET_SAVE');
+    });
   });
 
   it('should show error when isOwnerValid is false', async () => {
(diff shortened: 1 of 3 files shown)


SanzharBerdikozhoev:
In `kadai.date.spec.ts` setTimeout could not be easily replaced or deleted. 
That would require using of `fakeTimers` from vitest library, however they fatally slow down the test.

bruderj15 (commented):
Looks good.

> That would require using of fakeTimers from vitest library, however they fatally slow down the test.

`vi.waitFor` has a default timeout of 1000ms.
Afaics `kadai.date.spec.ts` uses `setTimeout(resolve, 1001);`
So maybe we can use `vi.waitFor` with a non-default timeout value (- maybe `1001`)?

Apart from that - although having never used it - I doubt that _they fatally slow down the test_ if correctly used.

bruderj15:
PS: Make sure you're not just ticking the PR-Description check-boxes but actually respect them (e.g. add missing linked issue to PR "Development" section).

SanzharBerdikozhoev:
Thank you for the feedback! 

As I have understood the timeout in `vi.waitFor` plays a role of a maximal time-limit, `vi.waitFor` can wait till the callback will be successfully executed (otherwise it will throw the error). But `vi.waitFor` checks the callback every 20 ms (default value) and if the callback has been run successfully (without any error), `vi.waitFor`  will stop its execution even before the timeout was reached. 

That's why the tests ran faster with `vi.waitFor` rather than with `Promise` with `setTimeout`.

1. With timeout test-result-with-promise
2. With waitFortest-result-with-waitFor

However it's not suitable for `kadai.date.spec.ts`-test, because there should be a callback that stops further test execution for at least 1001ms, which we have already in form of `Promise` with `setTimeout`. I have tried to use `waitFor` with `{timeout: 1001}`-parameter but it didn't help to solve the problem. 

```ts
it('should return different values on successive calls as time passes', async () => {
    const first = KadaiDate.getDate();

    // await new Promise((resolve) => setTimeout(resolve, 1001));

    // in my opinion that doesn't make sense
    await vi.waitFor(
      () => new Promise((resolve) => setTimeout(resolve, 1001)), {
      timeout: 1001,
    });

    const second = KadaiDate.getDate();
    expect(first).not.toBe(second);
  });
```

The only way i see now, is using of fake timers. I will try them one more time.

bruderj15:
I thought of:

```ts
it('should return different values on successive calls as time passes', async () => {
    const first = KadaiDate.getDate();
    await vi.waitFor(
      () => expect(first).not.toBe(KadaiDate.getDate()), {
      timeout: 2000, // technically 1021 would be sufficient afaics
    });
  }
);
```

SanzharBerdikozhoev:
oh, haven't think about it. It actually works. Thanks!

bruderj15 (approved):
Good job :)

PR #2131, opened 2026-08-21
Closes #2114 - Revisit Frontend TODOs

### LINKED ISSUE #2114: Revisit Frontend TODOs
Follow-Up to #2110

There are plenty of `TODO` and `FIXME` comments in `web/**`.
Revisit and handle them.

SanzharBerdikozhoev:
## Definition of Done

  - will be squashed on merge


web/src/app/shared/services/classifications/classifications.service.ts
SanzharBerdikozhoev: It seems to me it's not a good idea to change the behavior of existed function, especially which is widely-used like the service. However it's the most simple way to complete this TODO:

// TODO: this is currently called for every selected task and is only necessary when we switch the workbasket -> can be optimized.
private getClassificationByDomain() {
    this.requestInProgress.set(true);
    this.classificationService
      .getClassifications({ domain: [this.task()!.workbasketSummary!.domain!] })
      .pipe(takeUntil(this.destroy$))
      .subscribe((classificationPagingList) => {
        this.classifications.set(classificationPagingList.classifications);
        this.requestInProgress.set(false);
      });
  }

web/src/app/shared/models/classification.ts
```diff
@@ -21,8 +21,8 @@ import { Links } from './links';
 
 export interface Classification extends ClassificationSummary {
   isValidInDomain?: boolean;
-  created?: string; // TODO: make this a Date
-  modified?: string; // TODO: make this a Date
+  created?: Date;
+  modified?: Date;
```
bruderj15: `HttpClient` won't deserialize ISO timestamp strings into `Date` instances, so REST-loaded classifications will still contain strings at runtime. 
Could we either keep these as `string` or explicitly map `created`/`modified` with `new Date(...)` in ClassificationsService?

web/src/app/shared/store/classification-store/classification.state.ts
```diff
@@ -291,6 +293,6 @@ export interface ClassificationStateModel {
   classifications: ClassificationSummary[];
   selectedClassification: Classification;
   selectedClassificationType: string;
-  classificationTypes: CategoriesResponse;
+  classificationTypes: CategoriesMap;
```
bruderj15: Could we keep this state JSON-serializable? 
`Map` does not survive normal Redux DevTools serialization/time travel, while the selectors now depend on `.get()`/`.keys()`. 
A plain object here would avoid that issue.

SanzharBerdikozhoev: Good point! I initially thought first it is just REST request overhead, but breaking normal lifecycle of state manager makes it even worse. At least for now I see no crucial advantages over plain object. So I revert the changes.

web/src/app/shared/models/classification.ts
SanzharBerdikozhoev: Got it! I consider if we don't have a lot of date calculations and the dates are mostly to be read. Then it makes sense to leave them in format of `string`.

bruderj15 (approved):
Well done! :)

PR #2170, opened 2026-08-25
Closes #2132 - Reopen completed or cancelled tasks

### LINKED ISSUE #2132: Reopen completed or cancelled tasks
## Description
The backend exposes task reopening, but the Angular service/store/UI has no action and does not model or filter the `reopened` flag.

## REST API
- `POST /api/v1/tasks/{taskId}/reopen` → updated task
- `GET /api/v1/tasks?...&is-reopened=true|false`

## Definition of done
- [x] Add `reopenTask` to `TaskService` and an NGXS action/state handler.
- [x] Show **Reopen** only for `COMPLETED` and `CANCELLED`; never for final state `TERMINATED`.
- [x] Require confirmation, use the returned task as source of truth, and refresh the list/detail.
- [x] After success the UI reflects `CLAIMED`, current-user ownership, cleared `completed`, unread state, and `reopened = true`.
- [x] Add `reopened?: boolean` to the task model and an `is-reopened` task filter.
- [x] Surface authorization, invalid-state, not-found, concurrency, and callback-related backend errors without leaving stale state.
- [x] Cover service/store and component behavior, including both valid states and rejected `TERMINATED`/callback cases.

## References
- Backend change: [reopen commit](https://github.com/kadai-io/kadai/commit/308f2d672e3bc0a8c0a12b110a99c17985df04c1)
- Backend: [TaskController.java](https://github.com/kadai-io/kadai/blob/6308d9eef65744e4cbbfc327e456641a91a42410/rest/kadai-rest-spring/src/main/java/io/kadai/task/rest/TaskController.java), [ReopenTaskAccTest.java](https://github.com/kadai-io/kadai/blob/6308d9eef65744e4cbbfc327e456641a91a42410/lib/kadai-core-test/src/test/java/acceptance/task/reopen/ReopenTaskAccTest.java)
- Frontend gap: [task.service.ts](https://github.com/kadai-io/kadai/blob/6308d9eef65744e4cbbfc327e456641a91a42410/web/src/app/workplace/services/task.service.ts), [task.ts](https://github.com/kadai-io/kadai/blob/6308d9eef65744e4cbbfc327e456641a91a42410/web/src/app/workplace/models/task.ts)

SanzharBerdikozhoev:
## Definition of Done

  - will be squashed on merge
  - Added the ability to reopen `COMPLETED` and `CANCELLED` tasks, restoring them to `CLAIMED` status for the current user.
  - Introduced a new `is-reopened` filter to easily search for and identify reopened tasks.


web/src/app/shared/store/task-store/task.state.ts
```diff
@@ -259,6 +260,21 @@ export class TaskWorkflowState {
...
+  @Action(ReopenTask)
+  reopenTask(ctx: StateContext<TaskWorkflowStateModel>, action: ReopenTask): Observable<any> {
+    return this.withRequestInProgress(this.taskService.reopenTask(action.taskId)).pipe(
+      tap((task) => {
+        ctx.patchState({ selectedTask: task });
+        this.notificationService.showSuccess('TASK_REOPEN', { taskName: task.name });
+        ctx.dispatch(new LoadTasks());
+      }),
+      catchError((error) => {
+        this.notificationService.showError(error);
```
bruderj15: I don't think we should pass the whole HTTP error to `showError` here. `NotificationService.showError()` expects an error-key string, while the global HTTP interceptor already extracts `error.error.error.key` and `messageVariables` and displays it.

For non-404 errors this can therefore produce a second fallback/incorrect toast. More importantly, the interceptor intentionally skips 404s, so a reopen `TASK_NOT_FOUND` ends up here but loses its actual error key.

Could we either rely on the interceptor consistently, or explicitly extract `error.error.error.key` / `messageVariables` here for the cases this action needs to handle?

web/src/app/shared/services/obtain-message/message-by-error-code.ts
bruderj15, line 1: Could we also add the reopen-specific backend error (`REOPEN_TASK_WITH_CALLBACK`) to the `ERROR` section?

web/src/app/workplace/components/task-processing/task-processing.component.ts
bruderj15, line 86: When opening a `COMPLETED` or `CANCELLED` task, we first load it and then unconditionally dispatch `ClaimTask`. Is that intended for the new reopen flow?

These states transition back to `CLAIMED` via reopen, so attempting to claim them first should result in an invalid-state response. This means simply opening a task in order to reopen it produces an unnecessary error, and `loadAndClaimTask()` stops at this failed await.

May we skip `ClaimTask` when the loaded task is `COMPLETED` or `CANCELLED`?

web/src/app/workplace/components/task-processing/task-processing.component.spec.ts
```diff
@@ -291,6 +294,91 @@ describe('TaskProcessingComponent', () => {
...
+        callback();
+      });
+
+      component.reopenTask();
+
+      expect(dispatchSpy).not.toHaveBeenCalled();
+    });
+
+    it('should call reopenTask when reopen button is clicked', async () => {
+      await selectTask(makeTask({ state: 'COMPLETED' }));
```
bruderj15, line 349: Follow-Up to above comment:

Could we also add a test for opening/loading a `COMPLETED` or `CANCELLED` task through `loadAndClaimTask()` and verify that `claimTask` is not called? The current Reopen-button test sets a completed task directly in the store, so it doesn't cover the initialization path.

web/src/app/shared/services/obtain-message/message-by-error-code.ts
SanzharBerdikozhoev: Sure!

web/src/app/shared/store/task-store/task.state.ts
SanzharBerdikozhoev: Understood! I think we can delegate error handling to the interceptor, since other actions are also doing it.

web/src/app/workplace/components/task-processing/task-processing.component.spec.ts
SanzharBerdikozhoev: Sure!

web/src/app/workplace/components/task-processing/task-processing.component.ts
SanzharBerdikozhoev: Good Catch! Yes, definitely `ClaimTask` should be skipped when the task is `COMPLETED` or `CANCELLED`. I will add a condition for that.

web/src/app/workplace/components/task-processing/task-processing.component.spec.ts
```diff
@@ -291,6 +294,113 @@ describe('TaskProcessingComponent', () => {
     });
   });
 
+  describe('reopenTask()', () => {
```
bruderj15, line 300: The DoD contains `"Show Reopen only for COMPLETED and CANCELLED; never for final state TERMINATED."`.
Can you add a test for the `TERMINATED` condition?

NB: In general the DoD points are acceptance-criteria, so at least one test per criteria is recommended

SanzharBerdikozhoev: Thanks for the note! I have added two tests that check if `ClaimTask` action is called for the task with `TERMINATED` state (shouldn't) and if markup has `Reopen` button for the same task respectively.

bruderj15 (approved):

bruderj15:
I am so sorry. I did accidentally merge. Hope you can forgive me :(

SanzharBerdikozhoev:
No problem! :)

PR #2173, opened 2026-08-26
Closes #2166 - Prevent stale access id lookup responses in type ahead

### LINKED ISSUE #2166: Prevent stale Access-ID lookup responses in TypeAhead
Follow-Up to #2164: 

## Description
`TypeAheadComponent` starts an independent request for every debounced Access-ID value.

If lookup A is slow and lookup B completes first, A can still complete afterwards and overwrite the name, validity and emitted Access-ID with stale data.

## Proposed change
Use `switchMap` (or equivalent cancellation/latest-request handling) in the `valueChanges` pipeline so only the latest lookup may update component/parent state.

## Definition of Done
- [ ] Older Access-ID requests cannot overwrite newer input.
- [ ] Only the current input may emit `accessIdEventEmitter` / validity.
- [ ] Add a test with requests A → B resolving B → A and assert B remains selected.

SanzharBerdikozhoev:
## Definition of Done

  - will be squashed on merge
    - Fixed a race condition in the Access ID lookup. Outdated server responses no longer overwrite the latest input value.

### DIFF
--- a/web/src/app/shared/components/type-ahead/type-ahead.component.spec.ts
+++ b/web/src/app/shared/components/type-ahead/type-ahead.component.spec.ts
@@ -20,7 +20,7 @@ import { ComponentFixture, TestBed } from '@angular/core/testing';
 import { DebugElement } from '@angular/core';
 import { TypeAheadComponent } from './type-ahead.component';
 import { AccessIdsService } from '../../services/access-ids/access-ids.service';
-import { delay, of } from 'rxjs';
+import { delay, of, Subject } from 'rxjs';
 import { provideStore, Store } from '@ngxs/store';
 import { EngineConfigurationState } from '../../store/engine-configuration-store/engine-configuration.state';
 import { engineConfigurationMock } from '../../store/mock-data/mock-store';
@@ -189,11 +189,16 @@ describe('TypeAheadComponent with AccessId input', () => {
   });
 
   it('should ignore stale/older search requests if a new input was provided (prevent race condition)', async () => {
+    vi.useFakeTimers();
+
+    const requestA$ = new Subject<any>();
+    const requestB$ = new Subject<any>();
+
     const searchSpy = vi
       .spyOn(accessIdService, 'searchForAccessId')
       .mockReset()
-      .mockReturnValueOnce(of([{ accessId: 'user-a', name: 'User A' }]).pipe(delay(50))) // simulation of a long request
-      .mockReturnValueOnce(of([{ accessId: 'user-b', name: 'User B' }]).pipe(delay(10))); // the second request was sent later but was processed first
+      .mockReturnValueOnce(requestA$.asObservable())
+      .mockReturnValueOnce(requestB$.asObservable());
 
     const inputEl: HTMLInputElement = fixture.nativeElement.querySelector('input');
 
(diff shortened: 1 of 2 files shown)


web/src/app/shared/components/type-ahead/type-ahead.component.ts
```diff
@@ -107,14 +107,22 @@ export class TypeAheadComponent implements OnInit, OnDestroy {
...
-      .pipe(debounceTime(this.debounceTime), distinctUntilChanged(), takeUntil(this.destroy$))
-      .subscribe(() => {
-        const value = this.accessIdForm.controls['accessId'].value;
-        if (value === '') {
-          this.handleEmptyAccessId();
-          return;
-        }
-        this.searchForAccessId(value!);
+      .pipe(
+        debounceTime(this.debounceTime),
```
bruderj15: This does not fully satisfy the DoD. 
Because `debounceTime` is before `switchMap`, a newly entered value does not cancel the previous lookup immediately. While the new value is still being debounced, the previous request can complete and update `name`, validity and `accessIdEventEmitter`, even though the form control already contains the newer value.

Could we make the debounce part of the cancellable inner stream instead? E.g. let every raw `valueChanges` emission reach `switchMap` immediately, and debounce the lookup inside it. That way typing B immediately unsubscribes A while B's actual request remains debounced.

web/src/app/shared/components/type-ahead/type-ahead.component.ts
```diff
@@ -107,14 +107,22 @@ export class TypeAheadComponent implements OnInit, OnDestroy {
...
+        distinctUntilChanged(),
+        switchMap((value) => {
+          if (!value) {
+            this.handleEmptyAccessId();
+            return of(null);
+          }
+
+          return this.accessIdService.searchForAccessId(value).pipe(map((accessIds) => ({ value, accessIds })));
```
bruderj15: Please handle lookup errors inside this inner observable. With the lookup now being part of the main `valueChanges` pipeline, an HTTP error will propagate through `switchMap` and terminate the whole subscription. After one failed request, further typing would therefore no longer trigger lookups until the component is recreated.

Previously each lookup had its own subscription, so a failed request did not kill `valueChanges`. An inner `catchError(...)` (for example returning `EMPTY`, after any required error handling) would preserve that behavior.

web/src/app/shared/components/type-ahead/type-ahead.component.spec.ts
```diff
@@ -164,6 +164,69 @@ describe('TypeAheadComponent with AccessId input', () => {
...
+    });
+  });
+
+  it('should ignore stale/older search requests if a new input was provided (prevent race condition)', async () => {
+    const searchSpy = vi
+      .spyOn(accessIdService, 'searchForAccessId')
+      .mockReset()
+      .mockReturnValueOnce(of([{ accessId: 'user-a', name: 'User A' }]).pipe(delay(50))) // simulation of a long request
+      .mockReturnValueOnce(of([{ accessId: 'user-b', name: 'User B' }]).pipe(delay(10))); // the second request was sent later but was processed first
+
```
bruderj15, line 202: This test does not deterministically prove the DoD scenario “A → B resolving B → A and assert B remains selected”.

The configured debounce in this test setup is already 50 ms. A is delayed by only 50 ms, while B cannot even start until its own debounce has elapsed, so A may complete before B rather than after it. The test also finishes as soon as `name()` becomes `User B`; it never proves that B is still selected after A definitely resolves.

Could we use controllable `Subject`s (and preferably fake timers) instead: start A, start B, emit B, assert B's name/emitted AccessId/validity, then emit A afterwards and assert that none of those values changed? That would test the exact DoD without relying on wall-clock timing.

web/src/app/shared/components/type-ahead/type-ahead.component.ts
SanzharBerdikozhoev: Good Point! I will try to do this.

web/src/app/shared/components/type-ahead/type-ahead.component.ts
SanzharBerdikozhoev: Ok, I have got it. I will restore the error handling.

web/src/app/shared/components/type-ahead/type-ahead.component.spec.ts
SanzharBerdikozhoev: I have just tried using `Subject`s and fake timers one more time and it worked. I think this happened after moving `debounceTime` inside the `switchMap`. 

P.S. I also consider this approach is better, at the very least it doesn't slow down the test and not flaky, so I have kept it.

bruderj15 (approved):
Not gonna steal it this time :)

PR #2183, opened 2026-08-28
Closes #2167 - Prevent stale task load/claim pipelines on rapid route changes

### LINKED ISSUE #2167: Prevent stale task load/claim pipelines on rapid route changes
Follow-Up to #2164: 

## Description
`TaskProcessingComponent.loadAndClaimTask()` awaits `GetTask`, then `ClaimTask`, and afterwards reads the global `selectedTask`.

When route IDs change quickly, multiple pipelines can overlap. A late response for the previous task may overwrite state or cause the older pipeline to continue using the newer selected task.

## Proposed change
Make route changes cancel the previous load/claim workflow, e.g. with `switchMap`, and carry the task belonging to that workflow instead of rereading global mutable state.

## Definition of Done
- [x] Changing task A → B cancels/ignores A's unfinished workflow.
- [x] A late response for A cannot modify B's processing view/state.
- [x] Add a deterministic reversed-response-order test.

SanzharBerdikozhoev:
## Definition of Done

  - will be squashed on merge
    - Fixed a race condition on rapid route changes where stale task load/claim responses overwrote current data.


web/src/app/workplace/components/task-processing/task-processing.component.ts
```diff
@@ -73,29 +74,38 @@ export class TaskProcessingComponent implements OnInit, OnDestroy {
...
-    this.getWorkbaskets();
+  loadAndClaimTask(id: string): Observable<{ task: Task; classification: Classification }> {
+    return this.store.dispatch(new GetTask(id)).pipe(
+      switchMap(() => {
+        if (this.canClaimTask()) {
+          return this.store.dispatch(new ClaimTask(id));
+        }
+        return of(null);
+      }),
+      map(() => this.store.selectSnapshot(TaskSelectors.getSelectedTask)!),
```
bruderj15: Still depends on global mutable `selectedTask`, so race is not fully fixed. Outer `switchMap` unsubscribes from the `dispatch()` result, but doesn't cancel the NGXS action handler itself such that a late `GetTask(A)`/`ClaimTask(A)` can still patch `selectedTask` after B has loaded.

Could we make the state-changing workflow itself cancellable/guarded and carry the task belonging to this request instead of rereading `selectedTask`?

web/src/app/workplace/components/task-processing/task-processing.component.spec.ts
```diff
@@ -662,6 +662,72 @@ describe('TaskProcessingComponent', () => {
...
+      const claimTaskA$ = new Subject<void>();
+      const claimTaskB$ = new Subject<void>();
+
+      const classificationA$ = new Subject<any>();
+      const classificationB$ = new Subject<any>();
+
+      // allow claim logic to proceed through the switchMap chain
+      vi.spyOn(component, 'canClaimTask').mockReturnValue(true);
+
+      vi.spyOn(store, 'dispatch').mockImplementation((action: any) => {
```
bruderj15: Test mocks away the behavior we need to verify. In production, the real NGXS action handlers update `selectedTask` when their responses complete; these Subjects have no such side effect, and `selectSnapshot` is later forced to return B.

Could we keep the real Store/TaskWorkflowState and instead return controllable Subjects from the mocked `TaskService.getTask`/`claimTask`? Then complete B before A and assert that `selectedTask`, `component.task()` and the processing URL all remain B.

web/src/app/workplace/components/task-processing/task-processing.component.ts
```diff
@@ -73,29 +74,38 @@ export class TaskProcessingComponent implements OnInit, OnDestroy {
...
   ngOnInit() {
-    this.routeSubscription = this.route.params.subscribe((params) => {
-      const { id } = params;
-      this.loadAndClaimTask(id);
-    });
+    this.routeSubscription = this.route.params
+      .pipe(
+        map((params) => params['id']),
+        distinctUntilChanged(),
+        switchMap((id) => (id ? this.loadAndClaimTask(id) : EMPTY)),
```
bruderj15: An error from `GetTask`, `ClaimTask` or `getClassification` will now error this whole route-params stream, so subsequent route changes are no longer processed.

Could we catch errors inside the per-task inner workflow so a failed task load does not permanently terminate route handling?

SanzharBerdikozhoev: Sure, I will add an error handling inside the stream.

web/src/app/workplace/components/task-processing/task-processing.component.spec.ts
SanzharBerdikozhoev: Good Catch! Will not mock state actions and instead the methods of service.

web/src/app/workplace/components/task-processing/task-processing.component.ts
SanzharBerdikozhoev: Haven't thought about that, i.e. global state mutation. 
Of course will fix this. 
So will carry the task through the whole stream and as I have understood the easiest way to add protection is to filter the task by id and compare it with the current one to identify the same task. 
So, I will implement it this way.

web/src/app/workplace/components/task-processing/task-processing.component.ts
```diff
@@ -73,29 +87,51 @@ export class TaskProcessingComponent implements OnInit, OnDestroy {
...
   }
 
-  async loadAndClaimTask(id: string) {
-    await firstValueFrom(this.store.dispatch(new GetTask(id)));
-
-    const task = this.store.selectSnapshot(TaskSelectors.getSelectedTask)!;
+  loadAndClaimTask(id: string): Observable<{ task: Task; classification: Classification }> {
+    return this.store.dispatch(new GetTask(id)).pipe(
+      switchMap(() => this.store.select(TaskSelectors.getSelectedTask)),
+      filter((task): task is Task => !!task && task.taskId === id),
```
bruderj15: This filter happens after `GetTask` has already completed, so it cannot prevent the NGXS handler from patching global `selectedTask`, right?. A late A response may therefore still change `component.task()` back to A after B is active.

We need to prevent/guard the stale state update itself, not only ignore the stale component pipeline afterwards.

web/src/app/workplace/components/task-processing/task-processing.component.spec.ts
```diff
@@ -662,6 +664,80 @@ describe('TaskProcessingComponent', () => {
...
+      claimTaskA$.complete();
+
+      classificationA$.next({ applicationEntryPoint: 'http://app-a.com' } as Classification);
+      classificationA$.complete();
+
+      vi.runAllTimers();
+      fixture.detectChanges();
+
+      expect(component.address).toBe('http://app-b.com');
+
```
bruderj15, line 746: Could we also assert that the selected task is still B here?

```suggestion
      expect(component.address).toBe('http://app-b.com');
      expect(store.selectSnapshot(TaskSelectors.getSelectedTask)?.taskId).toBe('task-b');
      expect(component.task()?.taskId).toBe('task-b');
```

With the current implementation a late `GetTask(A)` can still overwrite global state while this address assertion continues to pass.

SanzharBerdikozhoev: Have added this asserts. Thanks!

web/src/app/workplace/components/task-processing/task-processing.component.ts
SanzharBerdikozhoev: I have added `activeTaskId` into `TaskWorkflowStateModel` to check the id of the actual task, that is being processed. I hope this will work.

web/src/app/shared/store/task-store/task.state.ts
```diff
@@ -173,13 +173,19 @@ export class TaskWorkflowState {
...
       this.requestInProgressService.setRequestInProgress(true);
       const newTask = new Task('', new ObjectReference(), ctx.getState().selectedWorkbasket);
-      ctx.patchState({ selectedTask: newTask });
+
+      ctx.patchState({ selectedTask: newTask, activeTaskId: 'new-task' });
       this.requestInProgressService.setRequestInProgress(false);
       return of(null);
     }
+
+    ctx.patchState({ activeTaskId: action.taskId });
```
bruderj15, line 184: This protects `selectedTask` but the stale request still reaches `withRequestInProgress()` and its `finalize()` emits `false`.

So A can finish while B is still pending and clear B's global loading indicator. Could we make the loading state safe for overlapping requests as well?

web/src/app/workplace/components/task-processing/task-processing.component.spec.ts
```diff
@@ -662,6 +664,82 @@ describe('TaskProcessingComponent', () => {
...
+      getTaskA$.next(taskA);
+      getTaskA$.complete();
+
+      claimTaskA$.next(taskA);
```
bruderj15, line 733: `claimTaskA$` has no subscriber here: we switch to B before `getTaskA$` completes, so A never reaches `ClaimTask`.

Could we add a case with: Get A completes → Claim A starts → switch to B → B completes → Claim A completes last, then assert state/address remain B?

bruderj15 (commented):
This is a tricky issue, almost there :)

web/src/app/workplace/components/task-processing/task-processing.component.spec.ts
SanzharBerdikozhoev: Have moved `getTaskA$` up before starting request `B`.

web/src/app/shared/store/task-store/task.state.ts
SanzharBerdikozhoev: I have tried to fix this with adding `defer()` inside `withRequestInProgress()`.

bruderj15: `defer()` only delays when `true` is emitted; it does not solve overlapping requests. A can still finalize and emit `false` while B is pending.

Could we make `requestInProgress` overlap-safe, e.g. via reference counting or a request/token guard?

SanzharBerdikozhoev: I will try reference counting.

bruderj15: The reference counting is almost right, but this unconditional `this.requestInProgressSubject.next(value)` still defeats it: when A finishes while B is active, it emits `false` anyway.

Could we remove that unconditional emit and add an intermediate test assertion that the value stays `true` after the first overlapping request finishes?

Then we got it!

SanzharBerdikozhoev: Sure!

bruderj15: Success-path overlap looks fixed. 
One error-path issue remains though: HTTP errors already call `setRequestInProgress(false)` in global interceptor, `withRequestInProgress()` then calls it again in `finalize()`.

With A+B active, an error in A can therefore decrement twice and hide B's spinner. Could we ensure each tracked request decrements exactly once and add an overlapping-error test?

SanzharBerdikozhoev: I have removed `setRequestInProgress(false)` call in global interceptor and added an overlapping-error test.

bruderj15: Task-specific overlap is fixed, but changing `setRequestInProgress(true/false)` into global reference counter breaks existing callers using it as a plain setter (e.g. sidenav + page component can both call `true`, with only one matching `false`).

Could we preserve `setRequestInProgress()` setter semantics and introduce separate balanced `beginRequest()` / `endRequest()` tracking for the overlapping task workflow instead?

SanzharBerdikozhoev: It seems to me it's added! Should we integrate new `beginRequest() / endRequest()` tracking mechanism right now?

bruderj15: I completed it, enough torture :)

bruderj15 (approved):

PR #2199, opened 2026-09-03
Closes #2149 - Split forms validator service into composable typed validators

### LINKED ISSUE #2149: Split `FormsValidatorService` into composable typed validators
## Description

The singleton `FormsValidatorService` combines unrelated responsibilities: global submit state, synchronous form validation, remote access-ID lookup, permission rules, notifications, animation toggles, and overflow timers. Its main methods return `Promise<any>` created around manual subscriptions; lookup errors have no reliable completion path. Shared mutable maps and the global `formSubmitAttempt` also allow one form to affect another.

## Proposed change

Replace the service with small, typed building blocks:

- local submitted/touched state per form;
- `ValidatorFn`/`AsyncValidatorFn` directives for required fields and access-ID existence;
- a pure permission-dependency validator;
- control errors or a focused directive for length-limit feedback;
- notifications triggered by the submitting component, not by validators.

Template-driven forms may remain where appropriate; the goal is standard Angular validation contracts rather than a mandatory full form rewrite.

## Definition of done

- [x] `validateFormInformation` and `validateFormAccess` no longer wrap subscriptions in manual promises.
- [x] No global mutable submit-attempt flag is shared between forms.
- [x] Access-ID validation is typed, cancellable, and defines an error policy.
- [x] Permission dependencies return structured validation errors and retain current warnings/UI feedback.
- [x] Overflow feedback does not use a singleton mutable map or retained timer subscriptions.
- [x] `Map<any, ...>`, `ResponseOwner.valid: any`, and `Promise<any>` are removed from touched code.
- [x] Task, workbasket, classification, and access-item form tests cover valid, invalid, lookup-error, and repeated-submit cases.

## Main scope

- `web/src/app/shared/services/forms-validator/`
- Consumers in task, workbasket, classification, and access-item forms

_Reviewed against `master@6308d9eef65744e4cbbfc327e456641a91a42410`._

SanzharBerdikozhoev:
## Definition of Done

  - will be squashed on merge


web/src/app/administration/components/workbasket-information/workbasket-information.component.html
```diff
@@ -71,16 +87,25 @@ <h6 class="subheading"> General </h6>
...
-                       type="text">
-                @if (inputOverflowMap().get(owner.name)) {
+                       class="form-control" 
+                       id="wb-owner" 
+                       maxlength="128"
+                       name="workbasket.owner" 
+                       placeholder="Owner" 
+                       required
+                       type="text"
+                       kadaiAccessIdValidator
```
bruderj15, line 102: Looks like this should be `kadaiAccessIdExists` (with the directive imported into the component). `kadaiAccessIdValidator` doesn’t match the new selector, so this input currently only gets the required check and a nonexistent owner can pass.

web/src/app/administration/components/workbasket-access-items/workbasket-access-items.component.ts
```diff
@@ -303,7 +305,9 @@ export class WorkbasketAccessItemsComponent implements OnInit, OnDestroy, AfterV
   }
 
   setAccessItemsGroups(accessItems: WorkbasketAccessItems[]) {
-    const AccessItemsFormGroups = accessItems.map((accessItem) => this.formBuilder.group(accessItem));
+    const AccessItemsFormGroups = accessItems.map((accessItem) =>
+      this.formBuilder.group(accessItem, { validators: [permissionDependencyValidator] })
```
bruderj15, line 309: Could we wire the new async access-ID validation to these controls too? After removing `validateFormAccess`, `accessId` only gets `Validators.required`, so any non-empty nonexistent ID can pass in the plain-input path.

web/src/app/shared/validators/permission-dependency.validator.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  if (permOpen && (!permReadTasks || !permRead)) {
+    warnings.push('PERM_OPEN_MISSING_DEPENDING_PERMISSIONS');
+  }
+
+  if (permDistribute && (!permAppend || !permTransfer)) {
+    warnings.push('PERM_DISTRIBUTE_MISSING_DEPENDING_PERMISSIONS');
+  }
+
+  return warnings.length > 0 ? { permissionWarnings: warnings } : null;
```
bruderj15, line 28: Could we keep these permission conditions non-blocking? Returning them as `ValidationErrors` makes each access-item group invalid, which can disable/create-block saves and later report `OWNER_NOT_VALID`. Previously these were warnings only and saving still proceeded.

web/src/app/workplace/components/task-information/task-information.component.html
```diff
@@ -133,12 +143,16 @@
             <mat-form-field appearance="outline">
               <mat-label>Owner</mat-label>
               <label for="ts-owner"></label>
-              <input #owner="ngModel" (input)="validateInputOverflow(owner, 32)"
-                     [(ngModel)]="task.owner" id="ts-owner"
+              <input #owner="ngModel"
+                     #ownerOverflow="overflowFeedback"
+                     kadaiAccessIdExists
```
bruderj15, line 148: Could we surface `invalidAccessId` / `accessIdLookupError` here? The async validator can make the form invalid, but this branch has no field error or notification, so Save can appear to do nothing for a bad owner. Previously `OWNER_NOT_VALID` gave feedback.

web/src/app/workplace/components/task-information/task-information.component.ts
```diff
@@ -155,14 +148,20 @@ export class TaskInformationComponent implements OnInit, OnDestroy {
...
-      if (value && !this.isClassificationEmpty && this.isOwnerValid) {
-        this.formValid.emit(true);
-      }
-    });
+    this.isClassificationEmpty = !task?.classificationSummary;
+
+    const formGroup = form.control;
+    formGroup.markAllAsTouched();
+
+    if (formGroup.valid && !this.isClassificationEmpty && this.isOwnerValid) {
```
bruderj15, line 162: Could we wait for pending async validation before deciding? While the owner lookup is `PENDING`, `formGroup.valid` is false; a quick Save click is dropped, and a successful lookup later won’t emit `formValid` until another click.

web/src/app/workplace/components/task-details/task-details.component.html
```diff
@@ -35,7 +35,7 @@
                   matTooltip="Save Task">
             <span> Save </span>
             <mat-icon>save</mat-icon>
-          </button>
+          </button>м
```
bruderj15, line 38: Small typo: the trailing Cyrillic `м` will render next to the Save button. Could we remove it?

bruderj15 (commented):
Thanks for the focused refactor — the new validators/directives are a nice direction. I found a few behavior gaps around access-ID validation and warning semantics that seem worth addressing.

web/src/app/workplace/components/task-details/task-details.component.html
SanzharBerdikozhoev: Sure, deleted it!

web/src/app/administration/components/workbasket-information/workbasket-information.component.html
SanzharBerdikozhoev: Ohh, replaced it.

web/src/app/shared/validators/permission-dependency.validator.ts
SanzharBerdikozhoev: Fixed it. Now validator returns only warning messages.

web/src/app/workplace/components/task-information/task-information.component.html
SanzharBerdikozhoev: Added a check in `task-information.component.ts`. It shows `OWNER_NOT_VALID` error message if access-id is invalid.

web/src/app/administration/components/workbasket-access-items/workbasket-access-items.component.ts
SanzharBerdikozhoev: Sure, have added access-ID-validator.

web/src/app/workplace/components/task-information/task-information.component.ts
SanzharBerdikozhoev: Sure, have added check for form status and waiting if it is `PENDING`.

web/src/app/administration/components/workbasket-information/workbasket-information.component.ts
```diff
@@ -61,6 +60,9 @@ import { CdkTextareaAutosize } from '@angular/cdk/text-field';
 import { MapValuesPipe } from '../../../shared/pipes/map-values.pipe';
 import { RemoveNoneTypePipe } from '../../../shared/pipes/remove-empty-type.pipe';
 import { toSignal } from '@angular/core/rxjs-interop';
+import { OverflowFeedbackDirective } from 'app/shared/directives/overflow-feedback.directive';
+import { FormFieldSubmitDirective } from 'app/shared/directives/form-field-submit.directive';
+import { FormSubmitDirective } from 'app/shared/directives/form-submit.directive';
```
bruderj15, line 65: Small follow-up: `kadaiAccessIdExists` is now in the template, but `AccessIdExistsValidatorDirective` is not imported/added to this standalone component. So the attribute is currently inert and the owner lookup still isn't validated here.

bruderj15 (commented):
Thanks for addressing the first pass — most of the fixes are heading in the right direction. I found a few remaining async-validation setup/timing edge cases worth tightening up.

web/src/app/administration/components/workbasket-access-items/workbasket-access-items.component.ts
```diff
@@ -306,6 +312,7 @@ export class WorkbasketAccessItemsComponent implements OnInit, OnDestroy, AfterV
     const AccessItemsFormGroups = accessItems.map((accessItem) => this.formBuilder.group(accessItem));
     AccessItemsFormGroups.forEach((accessItemGroup) => {
       accessItemGroup.controls.accessId.setValidators(Validators.required);
+      accessItemGroup.controls.accessId.setAsyncValidators(accessIdExistsValidator(this.accessIdService));
```
bruderj15: Could we apply the same validator setup to `addAccessItem()` and recalculate after `set*Validators`? Runtime `setValidators`/`setAsyncValidators` doesn't recompute status by itself, and new rows only get `required`, so loaded/new IDs can remain valid until edited.

web/src/app/administration/components/workbasket-access-items/workbasket-access-items.component.ts
```diff
@@ -365,20 +371,18 @@ export class WorkbasketAccessItemsComponent implements OnInit, OnDestroy, AfterV
...
 
-    const shouldSaveWorkbasket = await this.formsValidatorService
-      .validateFormAccess(this.buildValidationForm(accessItems), this.toggleValidationAccessIdMap)
-      .then((isFormValid) => isFormValid)
-      .catch(() => false);
+    this.checkPermissionWarnings();
 
-    if (shouldSaveWorkbasket) {
-      this.onSave(accessItemsUrl, accessItems);
+    if (this.accessItemsGroups.invalid) {
```
bruderj15: Could we also handle `PENDING` here? With async validators, `invalid` is false while pending, so Save can reach `onSave` before the access-ID lookup finishes. The new invalid-ID test waits for stability before submitting, so it doesn't cover this race.

web/src/app/administration/components/workbasket-information/workbasket-information.component.ts
```diff
@@ -164,30 +158,22 @@ export class WorkbasketInformationComponent implements OnInit, OnDestroy {
...
-      });
-  }
+    const form = this.workbasketForm();
+    if (!form) {
+      return;
+    }
 
-  isFieldValid(field: string): boolean {
-    return this.formsValidatorService.isFieldValid(this.workbasketForm(), field);
+    if (form.valid && this.isOwnerValid) {
```
bruderj15: Once the async owner validator is wired in, the opposite pending race exists here: `form.valid` is false while the lookup is `PENDING`, so a quick Save shows `WORKBASKET_SAVE` and is dropped. Could we wait for a non-pending status like task information does?

web/src/app/workplace/components/task-information/task-information.component.ts
```diff
@@ -155,14 +150,42 @@ export class TaskInformationComponent implements OnInit, OnDestroy {
...
-    this.formsValidatorService.validateFormInformation(this.taskForm(), this.toggleValidationMap).then((value) => {
-      if (value && !this.isClassificationEmpty && this.isOwnerValid) {
-        this.formValid.emit(true);
+    this.isClassificationEmpty = !task?.classificationSummary;
+
+    const formGroup = form.control;
+    formGroup.markAllAsTouched();
+
+    if (formGroup.pending) {
+      formGroup.statusChanges
```
bruderj15, line 168: This fixes the single-click pending case nicely. One edge case: every Save click while `PENDING` creates another one-shot subscription; when validation becomes valid, each can emit `formValid`, so two quick clicks may trigger multiple saves. Could we guard/cancel the previous pending wait?

web/src/app/administration/components/workbasket-information/workbasket-information.component.ts
SanzharBerdikozhoev: Didn't notice it. Added an import.

web/src/app/administration/components/workbasket-access-items/workbasket-access-items.component.ts
SanzharBerdikozhoev: Fixed it, added new private method for convenience.

web/src/app/administration/components/workbasket-access-items/workbasket-access-items.component.ts
SanzharBerdikozhoev: Added waiting for `PENDING` status.

web/src/app/workplace/components/task-information/task-information.component.ts
SanzharBerdikozhoev: Yes! Added `Subject` prevent multiple saves.

web/src/app/administration/components/workbasket-information/workbasket-information.component.ts
SanzharBerdikozhoev: Sure!

web/src/app/shared/validators/access-id-exists.validator.ts
```diff
@@ -0,0 +1,24 @@
...
+export function accessIdExistsValidator(accessIdsService: AccessIdsService): AsyncValidatorFn {
+  return (control: AbstractControl): Observable<ValidationErrors | null> => {
+    if (!control.value) {
+      return of(null);
+    }
+
+    const value = typeof control.value === 'string' ? control.value : control.value.accessId;
+
+    return of(value).pipe(
+      switchMap((searchVal) => accessIdsService.searchForAccessId(searchVal)),
```
bruderj15: This looks like a merge-conflict resolution regression after #2220: current master switched **workbasket access-item** validation to the new exact `validateAccessId()` endpoint specifically to avoid LDAP search-limit/fuzzy-search issues, but this refactor now routes it back through `searchForAccessId()`. Since this validator is also used for task/workbasket owners (with different auth semantics), could we preserve the new exact validation specifically for access items rather than reverting that master fix?

web/src/app/administration/components/workbasket-access-items/workbasket-access-items.component.ts
```diff
@@ -365,20 +371,22 @@ export class WorkbasketAccessItemsComponent implements OnInit, OnDestroy, AfterV
...
-
     if (!accessItemsUrl) {
       return;
     }
 
-    const shouldSaveWorkbasket = await this.formsValidatorService
-      .validateFormAccess(this.buildValidationForm(accessItems), this.toggleValidationAccessIdMap)
-      .then((isFormValid) => isFormValid)
-      .catch(() => false);
+    if (this.accessItemsGroups.pending) {
```
bruderj15: Could we validate the `accessItems` argument here rather than the current `accessItemsGroups`? In the create/copy flow we save the captured snapshot, while `SelectWorkbasket` can already replace the live form before this runs. The old test explicitly covered “target access items replace the form during validation”; restoring that scenario would help prevent the snapshot from escaping validation.

web/src/app/administration/components/workbasket-information/workbasket-information.component.ts
```diff
@@ -163,31 +159,27 @@ export class WorkbasketInformationComponent implements OnInit, OnDestroy {
...
-        }
-      });
-  }
+    const form = this.workbasketForm();
+    if (!form) {
+      return;
+    }
+
+    if (form.pending) {
+      await firstValueFrom(form.statusChanges!.pipe(filter((status) => status !== 'PENDING')));
```
bruderj15: Could we guard/cancel repeated pending submissions here too? Two Save clicks while validation is `PENDING` create two `firstValueFrom` waiters, and both can resume into `onSave()` once validation completes. The same pattern exists in access-items; the task form already has a cancellation guard we could mirror.

web/src/app/administration/components/workbasket-information/workbasket-information.component.html
```diff
@@ -71,16 +87,25 @@ <h6 class="subheading"> General </h6>
...
+                       kadaiAccessIdExists
+                       kadaiFormFieldSubmit
+                       #ownerField="kadaiFormField">
+                @if (ownerOverflow.isOverflowed()) {
                   <div class="error">{{ lengthError }}</div>
                 }
-                <kadai-shared-field-error-display [displayError]="!isFieldValid('workbasket.owner')"
-                                                  [validationTrigger]="toggleValidationMap.get('workbasket.owner')"
+                <kadai-shared-field-error-display [displayError]="ownerField.hasError"
                                                   errorMessage="* Owner is required" />
```
bruderj15: Could we distinguish an invalid/non-resolvable owner from `required` here? A non-empty owner rejected by the async validator currently shows `* Owner is required`, and submit only raises generic `WORKBASKET_SAVE`. Previously owner validation surfaced `OWNER_NOT_VALID`; task information now preserves that feedback too.

bruderj15 (commented):
Thanks — the previous follow-ups are addressed. I found a few remaining validation edge cases, mainly around the newly merged exact Access-ID validation and create/copy async behavior.

web/src/app/administration/components/workbasket-information/workbasket-information.component.html
SanzharBerdikozhoev: Added a special getter to differ the errors.

web/src/app/administration/components/workbasket-information/workbasket-information.component.ts
SanzharBerdikozhoev: Yes, added a `boolean` flag for it.

web/src/app/shared/validators/access-id-exists.validator.ts
SanzharBerdikozhoev: Now access-id-validator uses new `validateAccessId()` instead of `searchForAccessId()`.

web/src/app/administration/components/workbasket-access-items/workbasket-access-items.component.ts
```diff
@@ -365,19 +372,36 @@ export class WorkbasketAccessItemsComponent implements OnInit, OnDestroy, AfterV
...
+    }
+    this.isSubmitting = true;
+
+    try {
+      let isValid = false;
+      const isTestingLiveForm = this.isAccessItemsMatchingForm(accessItems);
+
+      if (isTestingLiveForm) {
+        if (this.accessItemsGroups.pending) {
+          await firstValueFrom(this.accessItemsGroups.statusChanges.pipe(filter((status) => status !== 'PENDING')));
```
bruderj15: There is still a stale-value race here: `accessItems` is captured before waiting, but we wait for the **live form** to become valid. If the user changes an access ID while its lookup is pending, Angular can cancel/revalidate the new value; once that succeeds, we can still save the old captured value. Could we either always validate the captured snapshot, or re-check that the live form still matches `accessItems` after the await before saving? A test that changes the ID while validation is pending would cover this nicely.

web/src/app/administration/components/classification-details/classification-details.component.ts
```diff
@@ -140,30 +140,25 @@ export class ClassificationDetailsComponent implements OnInit, OnDestroy {
...
+    const form = this.classificationForm();
+    if (!form) {
+      return;
+    }
+
+    if (form.valid) {
+      this.onSave();
+    } else {
+      form.control.markAllAsTouched();
+      this.notificationsService.showError('CLASSIFICATION_SAVE');
```
bruderj15: `CLASSIFICATION_SAVE` is not configured in `message-by-error-code.ts`, so this shows the generic fallback error. Could we use the existing `EMPTY_FIELDS` message (which also preserves the previous behavior), or add a dedicated configured message if we want different wording?

bruderj15 (commented):
Two remaining validation edge cases from the latest pass:

web/src/app/administration/components/classification-details/classification-details.component.ts
SanzharBerdikozhoev: Changed to `EMPTY_FIELDS` for the moment.

web/src/app/administration/components/workbasket-access-items/workbasket-access-items.component.ts
SanzharBerdikozhoev: Added a validation of the captured snapshot and a suitable test.

web/src/app/administration/components/workbasket-access-items/workbasket-access-items.component.ts
```diff
@@ -346,38 +353,42 @@ export class WorkbasketAccessItemsComponent implements OnInit, OnDestroy, AfterV
...
-      this.onSave(accessItemsUrl, accessItems);
+    try {
+      let currentAccessItems = accessItems ?? this.snapshotAccessItems();
+      const isValid = await this.validateAccessItemsSnapshot(currentAccessItems);
+      if (!isValid) {
+        this.notificationsService.showError('OWNER_NOT_VALID', { owner: 'access id' });
+        return;
+      }
+      if (!accessItems) {
+        currentAccessItems = this.snapshotAccessItems();
```
bruderj15, line 387: This still allows validating one snapshot and saving another: we validate `currentAccessItems`, then for the normal live-form path replace it with a fresh snapshot without revalidating. If the user changes `USER_OLD` to an invalid `USER_NEW` while validation is pending, OLD can validate and NEW can then be saved unvalidated. Could we save the validated snapshot, or revalidate whenever the fresh snapshot differs? The new test currently uses one shared subject returning `true` for both values, so it doesn’t catch this invalid-new-value case.

web/src/app/administration/components/workbasket-access-items/workbasket-access-items.component.html
```diff
@@ -103,14 +103,15 @@
                   </td>
                 } @else {
                   <td>
-                    <div [class]="{ 'has-warning': (accessItemsClone()[index].accessId !== accessItem.value.accessId),
-																  'has-error': !accessItem.value.accessId && formsValidatorService.formSubmitAttempt}">
+                    <div [class.has-warning]="accessItemsClone()[index].accessId !== accessItem.value.accessId"
+                         [class.has-error]="accessIdField.hasError">
```
bruderj15, line 107: Small UX regression: Save is triggered from the parent toolbar, so this child form never emits its native submit event, and `onSubmit()` does not mark controls touched. That means an untouched invalid Access ID won’t get this row-level `has-error` state after Save anymore (the old `formSubmitAttempt` did). Could we mark the live form/controls touched on failed submit so the invalid row is highlighted too?

bruderj15 (commented):
Two small follow-ups remain from the latest validation pass:

web/src/app/administration/components/workbasket-access-items/workbasket-access-items.component.html
SanzharBerdikozhoev: Added `this.AccessItemsForm.markAllAsTouched();` if the form is invalid.

web/src/app/administration/components/workbasket-access-items/workbasket-access-items.component.ts
SanzharBerdikozhoev: Current snapshot is saved now, not replaced by the new one and test is updated.

bruderj15 (approved):
Here we go!

PR #2215, opened 2026-09-08
Closes #2168 - Validate Access-ID against the value that triggered the lookup

### LINKED ISSUE #2168: Validate Access-ID against the value that triggered the lookup
Follow-Up to #2164: 

## Description
`validateFormInformation()` starts an Access-ID lookup using the current owner value, but when the response arrives it rereads the live form value.

The form can therefore request validation for owner A but evaluate the result against owner B if the field changes meanwhile.

## Proposed change
Snapshot the owner value before starting the lookup and validate the response against that same value. Prefer cancellable Angular async validation as described in #2149.

## Definition of Done
- [x] Validation result always corresponds to the value that triggered the request.
- [x] Changing the owner during validation cannot produce a stale valid/invalid result.
- [x] Add a test where owner A is requested, changes to B, then A's response arrives.

SanzharBerdikozhoev:
## Definition of Done

  - will be squashed on merge
    - Fixed an issue where changing the owner field during async validation could accept stale responses for the previous value.

### DIFF
--- a/web/src/app/shared/services/forms-validator/forms-validator.service.spec.ts
+++ b/web/src/app/shared/services/forms-validator/forms-validator.service.spec.ts
@@ -22,7 +22,7 @@ import { FormsValidatorService } from './forms-validator.service';
 import { AccessIdsService } from 'app/shared/services/access-ids/access-ids.service';
 import { NotificationService } from '../notifications/notification.service';
 import { FormArray, FormControl } from '@angular/forms';
-import { of } from 'rxjs';
+import { of, Subject } from 'rxjs';
 
 const accessIdsServiceMock = {
   searchForAccessId: vi.fn().mockReturnValue(of([{ accessId: 'user1' }]))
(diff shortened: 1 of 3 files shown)


web/src/app/shared/services/forms-validator/forms-validator.service.ts
```diff
@@ -56,11 +56,17 @@ export class FormsValidatorService {
...
+      if (ownerControl) {
+        const requestedOwnerValue = ownerControl.value;
+
+        this.accessIdsService.searchForAccessId(requestedOwnerValue).subscribe((items) => {
           const validationState = toggleValidationMap.get(this.workbasketOwner);
           toggleValidationMap.set(this.workbasketOwner, !validationState);
-          const valid = items.find((item) => item.accessId === form.form.controls[this.workbasketOwner].value);
+          const stillCurrent = ownerControl.value === requestedOwnerValue;
+          const matches = items.some((item) => item.accessId === requestedOwnerValue);
+          const valid = stillCurrent && matches;
```
bruderj15: Stale responses still resolve `false`, triggering `OWNER_NOT_VALID`. Please ignore/cancel them before side effects.

web/src/app/shared/services/forms-validator/forms-validator.service.spec.ts
```diff
@@ -174,6 +174,69 @@ describe('FormsValidatorService', () => {
...
+        }
+      };
+
+      const toggleMap = new Map<any, boolean>();
+      const validationPromise = service.validateFormInformation(mockForm, toggleMap);
+      console.log(validationPromise);
+
+      mockOwnerValue = 'ownerB';
+
+      accessIdSubject.next([{ accessId: 'ownerA' }]);
```
bruderj15: This also passes on `master`. Return `ownerB` (or A+B) here so the old live-value comparison would incorrectly pass.

web/src/app/shared/services/forms-validator/forms-validator.service.spec.ts
```diff
@@ -174,6 +174,69 @@ describe('FormsValidatorService', () => {
...
+              invalid: false,
+              valid: true
+            }
+          }
+        }
+      };
+
+      const toggleMap = new Map<any, boolean>();
+      const validationPromise = service.validateFormInformation(mockForm, toggleMap);
+      console.log(validationPromise);
```
bruderj15: Please remove this debug log.

bruderj15 (commented):
Three small review notes.

web/src/app/shared/services/forms-validator/forms-validator.service.spec.ts
SanzharBerdikozhoev: Oh, sorry, haven't noticed.

web/src/app/shared/services/forms-validator/forms-validator.service.spec.ts
SanzharBerdikozhoev: Changed `ownerA` to `ownerB`.

web/src/app/shared/services/forms-validator/forms-validator.service.ts
SanzharBerdikozhoev: Added request relevance validation to prevent showing error notification.

bruderj15: This still becomes `false`, so `onSubmit()` shows `WORKBASKET_SAVE`. Please keep stale/cancelled distinct from invalid.

web/src/app/shared/services/forms-validator/forms-validator.service.spec.ts
```diff
@@ -16,13 +16,13 @@
...
 
-import { TestBed } from '@angular/core/testing';
+import { fakeAsync, TestBed, tick } from '@angular/core/testing';
 import { beforeEach, describe, expect, it, vi } from 'vitest';
 import { FormsValidatorService } from './forms-validator.service';
 import { AccessIdsService } from 'app/shared/services/access-ids/access-ids.service';
 import { NotificationService } from '../notifications/notification.service';
-import { FormArray, FormControl } from '@angular/forms';
-import { of } from 'rxjs';
+import { FormArray, FormControl, NgForm } from '@angular/forms';
```
bruderj15: `fakeAsync`, `tick`, and `NgForm` are unused; please remove them.

SanzharBerdikozhoev: Sure, my fault.

SanzharBerdikozhoev:
Tried to use pre-push github-hook to autoformat code and to check for unnecessary imports, debug logs etc. It was called but didn't notice the changes in commit, so will try to fix it :)

web/src/app/shared/services/forms-validator/forms-validator.service.ts
SanzharBerdikozhoev: Now `validateFormInformation` returns `null` if the request is stale. Also I have added check for `null` in `workbasket-information.component` so it should not show an error now if the owner control value was changed during the request. Could you check the changes please?

bruderj15 (approved):

PR #2216, opened 2026-09-08
Closes #2169 - Require exact Access-ID matches when validating workbasket access items

### LINKED ISSUE #2169: Require exact Access-ID matches when validating workbasket access items
Follow-Up to https://github.com/kadai-io/kadai/issues/2164:

## Description
`validateFormAccess()` currently treats `searchForAccessId()` as valid whenever the result list is non-empty.

The endpoint performs a search, so an input such as `user` may return `user-a` / `user-b` even though `user` itself is not a valid Access-ID.

## Proposed change
Require an exact Access-ID match in the returned results, preferably case-insensitive, consistent with `TypeAheadComponent`.

## Definition of Done
- [ ] A search result is valid only when it contains the entered Access-ID exactly.
- [ ] Partial/fuzzy-only matches are rejected.
- [ ] Add tests for exact match, case-insensitive exact match, and fuzzy-only results.

SanzharBerdikozhoev:
## Definition of Done

  - will be squashed on merge
    - Fixed workbasket access item validation to require an exact, case-insensitive Access-ID match and reject partial or fuzzy search results.

### DIFF
--- a/web/src/app/shared/services/forms-validator/forms-validator.service.spec.ts
+++ b/web/src/app/shared/services/forms-validator/forms-validator.service.spec.ts
@@ -202,12 +202,50 @@ describe('FormsValidatorService', () => {
       expect(result).toBe(true);
     });
 
+    it('should resolve to true when access ID matches case-insensitively', async () => {
+      accessIdsServiceMock.searchForAccessId.mockReturnValue(of([{ accessId: 'USER1' }]));
+
+      const formArray = new FormArray([
+        new FormControl({
+          accessId: 'user1',
+          permRead: true,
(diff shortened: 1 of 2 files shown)


bruderj15 (approved):
Looks good.
Took me a second though to notice that:
- we now forbid valid values
- when they were not part of the search-result page
- due to the currently awkward backend search-limit-50-then-sort limitation.

This is fine here, yet worth noting that I created #2217 as follow-up

SanzharBerdikozhoev:
Ohh, didn't know about this limitation! 
Good luck with #2217! 
And thanks for the review!

